PR #1, opened 2022-10-20
Include array STL header

tncowart:
Without this header the file wouldn't compile on macOS 12.6


bnprks:
I'm honestly confused how that was building without errors on my machine. Thanks for the fix!

PR #2, opened 2022-10-20
Fix failing ChrSelectRoundTrip test

tncowart:
The test was failing because it used a file that isn't checked in to the repository. Changing the file to one that is fixes the test.


bnprks:
Thanks for the fix! The repo has been mostly in single-contributor mode for a while, so I expect there's many of these one-off leftovers to clean up

PR #5, opened 2023-03-10
Add round method.

brgew:


bnprks:
Thanks for the contribution! I made a couple small edits which you can check out in the commit history. (I shoved a few unrelated cleanup bits in there as well, so I apologize for the function re-ordering that messes up the diff view)

The main changes were:
- Taking out all code related to digits != 0, since we won't support that initially
- Implementing a method of the `round()` S4 generic rather than overwriting it in the BPCells namespace
- Explicitly setting the rounding mode in C++ in case some user's code messes with it

PR #21, opened 2023-04-28
Add binarize method.

brgew:


R/transforms.R
```diff
@@ -257,6 +257,45 @@ min_by_col <- function(mat, vals) {
...
+  c(
+    short_description(x@matrix),
+    "Transform matrix to binary; that is, zeros and ones"
+  )
+})
+
+setMethod("binarize", "IterableMatrix", function(x, threshold=1) {
+  assert_is(x, "IterableMatrix")
+  assert_is(threshold, "numeric")
+  if (threshold < 0) stop("binarize threshold must be greater than or equal to zero")
```
bnprks, line 293: I think this should be `threshold <= 0`, since `threshold == 0` would change 0s to 1 which is not what we want.

bnprks:
This looks good overall, thanks for another contribution! There are a couple edits I'd request. Feel free to make as many or as little of these changes as you like, and I can tidy up the remainder. Just let me know your preference.

1. As per my code comment above, I think the error check should be at `threshold <= 0`
2. Add an option to do strict inequality (as in `<` as well as `<=`). This would probably work easiest by adding a second global parameter to the transform that is 0 or 1 on the C++ side (TRUE/FALSE on the R side)
3. Rather than (or in addition to) defining a `binarize` function, it would make sense to implement new generics for `<`, `<=`, `>`,  and `>=` [R docs link](https://web.mit.edu/r/current/lib/R/library/base/html/Comparison.html) (see section on S4 methods).  

We can just implement `mat < num` and `num > mat` (the directions that always can preserve sparsity). Optionally implement the other type combinations just to throw an informative error. The declaration lines should look something like this:
```
setMethod("<", signature(x= "IterableMatrix", y= "numeric"), function(x, y) {....})
setMethod(">", signature(x= "numeric", y= "IterableMatrix"), function(x, y) {....})
```
For now, also throw an error if the user passes a vector of length >1, since we'll only handle single numbers at this point.

Let me know if you're happy using the comparison operators instead of separately defining a `binarize` function -- I'm also open to keeping in a function called `binarize` since it's a bit more discoverable than the builtin comparison operators

bnprks:
Oh, I think that compile issue should be fixed as of about an hour ago. Eventually I'll set up automatic tests to prevent changes which work on my laptop but break for other operating systems. Until then these types of bug reports are very helpful for the occasional breaking change that sneaks in. (There is a separate "issues" feature on github which is a nice way to report problems)

In terms of matrix type, it might be nice to convert back to a `uint32_t` matrix after the transform, but definitely keep it numeric. It would even be possible to add an optimization so that `(mat >= threshold) + 0` works better by making `+ 0` into a no-op within BPCells

bnprks:
Hi Brent,

Thanks for the contribution! I think the strict_inequality parameter looks fine and the conversions should all work properly. I've made a couple cleanup edits described in [this commit](https://github.com/bnprks/BPCells/pull/21/commits/a0e7415dff7a58a27bd9a653fa1e3019dd9e88b2).

Otherwise, looks good and I'll merge it into main now

-Ben

PR #27, opened 2023-05-09
Add digits argument to round method.

brgew:


src/matrixTransforms/Round.cpp
```diff
@@ -12,10 +12,20 @@ bool Round::load() {
...
+    if(digits == 0) {
+      for (uint32_t i = 0; i < cap; i++) {
+          val_data[i] = std::nearbyint(val_data[i]);
+      }
     }
+    else {
+      const double factor = pow(10.0, (double)digits);
+      for (uint32_t i = 0; i < cap; i++) {
+        val_data[i] = std::nearbyint(val_data[i] * factor) / factor;
+      }
```
bnprks, line 26: Since division much slower than multiplication, we can do a little optimization here of pre-computing `inv_factor = 1/factor` in a separate variable, then doing `std::nearbyint(val_data[i] * factor) * inv_factor`

[This reference](https://www.agner.org/optimize/instruction_tables.pdf) lists multiplication (FMUL) at 1 cycle, and division (FDIV) at 4-5 cycles for 10th gen Intel chips. This will slightly mess with the precision of our operation

R/transforms.R
```diff
@@ -369,9 +369,9 @@ setMethod("short_description", "TransformRound", function(x) {
 setMethod("round", "IterableMatrix", function(x, digits=0) {
   assert_is(x, "IterableMatrix")
   assert_is(digits, "numeric")
-  if (digits != 0) stop("BPCells only supports round() with digits=0")
+  if (digits < 0) stop("Digits value must be greater than or equal to zero")
```
bnprks, line 372: Do we need this restriction? R allows `round(x, -1)` to round to the nearest 10, which I think would work seamlessly with the existing code you've written

R/transforms.R
```diff
@@ -354,13 +354,13 @@ setMethod(">=", signature(e1= "numeric", e2= "IterableMatrix"), function(e1, e2)
...
 setMethod("iterate_matrix", "TransformRound", function(x) {
-  iterate_matrix_round_cpp(iterate_matrix(x@matrix))
+  iterate_matrix_round_cpp(iterate_matrix(x@matrix), x@global_params[1])
 })
 
 setMethod("short_description", "TransformRound", function(x) {
   c(
     short_description(x@matrix),
-    "Transform round to nearest integer"
+    "Transform round to specified precision"
```
bnprks, line 363: Could we make this a bit more descriptive using `sprintf`?, like `sprintf("Transform round to %d decimal places", x@global_params[1])`

tests/testthat/test-matrix_transforms.R
```diff
@@ -99,13 +99,20 @@ test_that("binarize works", {
...
+        round(digits=digits) %>%
+        as("dgCMatrix")
+    expect_equal(round(m, digits=digits), res, tolerance=(10.0**-digits))
+
+    digits <- 2
+    res <- m %>%
+        as("IterableMatrix") %>%
+        round(digits=digits) %>%
+        as("dgCMatrix")
+    expect_equal(round(m, digits=digits), res, tolerance=(10.0**-digits))
```
bnprks, line 115: I don't think we want to set the `tolerance` argument so loosely. Using the default value (`testthat_tolerance() == 1.49e-08`) would be better

bnprks:
Yeah, we can adjust tolerance, but I'd like something that's at least as stringent as 1.5e-08.  My concern is that with the existing tolerance cutoffs, something like this would pass:
```R
digits <- 0
expect_equal(1.1, 1, tolerance=10.0**digits)
# I'd really like the test to complain if we're off by a whole 0.1
```

bnprks:
Great, thanks for making those changes! Merged into the main branch now

PR #70, opened 2024-01-29
checking data type before merging different matrices

Yunuuuu:
fix https://github.com/bnprks/BPCells/issues/68


Yunuuuu:
But it's hard to deal with `[<-` method since it use `dgCMatrix` class object as a substituted value, which only accept numeric data type, this means `[<-` cannot work for integer `IterableMatrix` with usual R matrix.

Yunuuuu:
One way is to always keep the data type of `x`, and use `convert_type` for `value` argument before proceeding.

bnprks:
Thanks for spotting the bug and suggesting a fix! I've added a regression test to cover the fixed error, and gone with your first suggested strategy for handling `[<-` (convert to match x, and output a warning if value gets automatically converted)

Issue #76, opened 2024-03-01
Try solving the issue "BPCells to H5AD"


ycli1995:
Hi, @bnprks. For issue https://github.com/bnprks/BPCells/issues/49#issuecomment-1932869295, when I force the `H5NumWriter<uint64_t>` to write `int64_t` into HDF5 dataset, the .h5ad file can work normally with python `sc.read_h5ad`. See the branch https://github.com/ycli1995/BPCells/tree/anndata.

Below is my example:
```
{r}
devtools::with_debug(devtools::load_all("/develop/ycli1995/BPCells"))

# open an old 10x HDF5 with specified 'group'
mat1 <- open_matrix_10x_hdf5("/develop/1M_neurons_filtered_gene_bc_matrices_h5.h5", group = "mm10")

matrix_type(mat1)  ## uint32_t

mat1 <- mat1[1:200, 1:100]
mat1 <- convert_matrix_type(mat1, "double")
mat1 <- write_matrix_anndata_hdf5(mat1, "sct_counts.h5")
```
```
{python}
>>> import anndata as ad
>>> adata = ad.read_h5ad("sct_counts.h5")
>>> adata[0:9, 0:9]
View of AnnData object with n_obs × n_vars = 9 × 9
    obs: 'bpcells_name'
    var: 'bpcells_name'

>>> adata[0:9, 0:9].X
<9x9 sparse matrix of type '<class 'numpy.float64'>'
        with 5 stored elements in Compressed Sparse Row format>

>>> adata.X.indptr
array([   0,   10,   17,   30,   42,   63,   91,  101,  117,  129,  147,
        169,  187,  195,  214,  238,  259,  279,  320,  341,  346,  366,
        389,  403,  413,  423,  432,  445,  458,  481,  509,  521,  529,
        558,  573,  587,  603,  623,  631,  641,  648,  655,  659,  679,
        693,  708,  714,  725,  736,  749,  762,  772,  779,  801,  814,
        826,  845,  852,  864,  880,  899,  908,  931,  944,  951,  963,
        973,  991, 1003, 1025, 1031, 1045, 1062, 1071, 1085, 1099, 1127,
       1139, 1153, 1172, 1193, 1227, 1257, 1273, 1281, 1295, 1311, 1320,
       1332, 1348, 1356, 1371, 1381, 1392, 1410, 1422, 1432, 1438, 1450,
       1461, 1468])

>>> adata.X.indices
[...]

Issue #49, opened 2023-09-26
BPCells to h5AD


jonathan-columbiau:
Hi,

Thanks for creating this great tool! I'd like to use data I currently have stored in a BPCells matrix with a library only found in Python and that takes in h5ad Anndata files - does BPCells have functionality to write matrices to h5ad?

Thanks!

bnprks:
Currently no, but I've been waiting for an excuse to add the functionality. I'll try adding it over the next couple days and post back here how it goes

bnprks:
Sorry for the delay on this, but I've finally pushed an update which adds the function `write_matrix_anndata_hdf5()`. This should make it possible to write a BPCells matrix either to a standalone h5ad file, or as an extra matrix in an existing h5ad

jonathan-columbiau:
Thanks, really appreciate it!

Dario-Rocha:
Hello again, this package is getting everyday better!
At the moment I have a similar task as the OP but I R fails to find the function

> Error: 'write_matrix_anndata_hdf5' is not an exported object from 'namespace:BPCells'

Even though the function is there in the help
I've just reinstalled BPCells package

Thanks a lot again for your help!

bnprks:
Oh thanks for the heads up @Dario-Rocha! It looks like I hadn't re-generated the NAMESPACE file so the function wasn't getting exported. Should be fixed now by commit f1b9f6bb. (Also for future temporary workarounds, you can use `BPCells:::` with three colons to access unexported methods though please still let me know if I've forgotten to export something)

Dario-Rocha:
Great! that worked. Now I am only wondering if BPCells can write the expression matrix along with the obs and vars data in the h5ad file

bnprks:
Right now the intended behavior is as follows (though correct me if you're seeing something else happen):
1. BPCells can write sparse matrices either to the main matrix `X` (by default), a layer of the matrix (by setting the group to `layers/my_matrix_name`), or even under `varm` or `obsm` if desired.
2. For compatibility purposes, BPCells will write a  barebones `obs` or `var` group if that doesn't already exist in the file, which just contains a 0-based row or column index.

Assuming you are wanting to write additional data to `obs` or `var`, BPCells doesn't have additional support for that right now. I'd be happy to accept a contribution adding this support, but I haven't built it personally because it seems a bit tricky to get right and BPCells doesn't have much other metadata-related functionality right now.

[...]

PR #81, opened 2024-03-13
[cpp]Force the `X/indptr` to be int64 for `createAnnDataMatrix`

ycli1995:
Solve https://github.com/bnprks/BPCells/issues/76


bnprks:
This looks great, thank you! Looks like a minimal set of changes that from my checks have fixed the issues. 

Also from my checks, I don't think we need to perform any conversions when reading because the HDF5 library appears to do basic conversions automatically (so if we open the array as `uint64_t` it doesn't matter if the underlying storage was actually `int64_t`)

I don't think there's an easy automated test we can add, but for posterity a quick way to test if this works (as previously described in #76 and #49 comment)
R code:
```r
library(BPCells)
library(Matrix)

m <- matrix(1:12, nrow=3) |> as("dgCMatrix") |> as("IterableMatrix")
write_matrix_anndata_hdf5(m, "test_matrix.h5ad")
```
Python code:
```py
import anndata as ad
adata = ad.read_h5ad( "test_matrix.h5ad")
adata.X[:2,:2]
```

In the old version, this will output `ValueError: unsupported data types in input`, and in the new version this will output the subset sparse matrix object.

PR #98, opened 2024-07-03
fix setMethod not found when loading

Yunuuuu:
When using BPCells in other package, R CMD Check report following errror:
image


bnprks:
Hi @Yunuuuu, thanks for noticing this and bringing it up! After seeing this pull request I did some more research, and I think there might actually be several more functions from the `methods` class that BPCells uses without either listing them in the `NAMESPACE` file or qualifying them with a `methods::` prefix. 

Do you mind if I make some additional changes to this pull request? I'll add everything we need into `NAMESPACE` so it's not necessary to add `methods::` prefixes -- only downside is it will add a few new commits to your main branch that you'll need to fetch.

Yunuuuu:
Please go ahead and make the necessary changes.

bnprks:
Okay, I've pushed a new change from my end. Could you run a pull and confirm that still solves the `setMethod` error from your end? I still have several more R CMD check warnings to work through, but I think this should solve that particular error.

Yunuuuu:
Only two functions in the `.on_load` will cause the error information. Therefore, the initial commit fixed the error message. I will now attempt the new commit, and thanks for your help.

Yunuuuu:
It works well in my test. No `.on_load` error message

bnprks:
Great, thanks @Yunuuuu! I'll merge this in and get to the other more minor CRAN warnings in separate pull requests.

PR #103, opened 2024-07-05
[r] add row/col max stat functions

immanuelazn:
# Description
Added row max and col max standalone rudimentary functions for iterable matrices.  Can handle transposed matrices.

# Tests
Tested against passing the non-iterable matrix version through MatrixStats and asserting equality
- matrices with no zeroes
- matrices with some, or all zeroes
- transposed matrices

# Thoughts
Would it make more sense to make this a method within IterableMatrix rather than a standalone function?  I see some functions in Matrix.R that aren't methods for their specific matrix class, but instead assert typing.  Not sure if that is the correct route for me to take though.


r/src/matrix_utils.cpp
```diff
@@ -624,6 +623,32 @@ List apply_matrix_double_cpp(SEXP mat_sexp, Function f, bool row_major) {
     return ret;
 }
 
+// Compute the max of each row
+// [[Rcpp::export]]
+NumericVector matrix_compute_max_per_row(SEXP matrix) {
```
bnprks, line 628: I'd suggest naming these `matrix_max_per_row_cpp` and `matrix_max_per_col_cpp`. I've been puting the `_cpp` ending on all Rcpp functions to make it clear where in the R code is calling into C++. And I think the "compute" part of the name is not adding a lot of information to the name (though that's just a matter of taste)

r/tests/testthat/test-matrix_stats.R
```diff
@@ -187,4 +187,62 @@ test_that("svds registers generic with RSpectra", {
...
\ No newline at end of file
+})
+
+test_that("row_max works for iterable matrices", {
+  withr::local_seed(195123)
+  m1 <- matrix(1, nrow = 3, ncol = 4)
+  i1 <- m1 %>% as("dgCMatrix") %>% as("IterableMatrix")
+  expect_equal(row_max(i1), matrixStats::rowMaxs(as.matrix(m1)))
+})
+
```
bnprks, line 198: Regarding the test cases:

1. The matrices are mostly initialized to constants, so using one of the helpers like `generate_sparse_matrix` or `generate_dense_matrix` might be useful to create slightly more complicated test cases.
2. One set of test areas that aren't addressed here is matrix rows/cols that are all negative, or are all negative/0. 

Finally, it might be nice to consolidate these tests a bit (though this is also a matter of personal preference). A recipe I like for these types of things is:

1. Make a hand-crafted test matrix that hits all the corner cases I'm worried about across its various rows.
2. Make a randomly-generated sparse matrix that's about 4 x 64, which should cover all sorts of sparsity patterns given that `generate_sparse_matrix` defaults to 50% sparsity.

Then for each of the matrices test that rowMaxs and colMaxs both work for the transpose and the original matrix (both transposing before and after converting to IterableMatrix). 
Putting those tests in a loop is an option if you have enough variants to test, though in this case it might not be needed.

My attitude on these tests is that the goal is to flag the presence of a problem but it's okay if it takes a bit more investigation to nail down the precise nature of the bug if it exists.

r/R/matrix.R
```diff
@@ -158,6 +157,36 @@ all_matrix_inputs <- function(x) {
   x
 }
 
+#' Get the max of each row in an iterable matrix
+#' @param x IterableMatrix object/dgCMatrix object
+#' @export
+row_max <- function(x) {
```
bnprks, line 163: For naming clarity, I think we'll want to do a similar approach as `rowVars()`/`colVars()` below, namely:
1. Define a `rowMaxs` generic function with the same arguments as `matrixStats::rowMaxs`
2. Define `rowMaxs.default` to call out to `matrixStats` or `matrixGenerics` if present
3. Put our actual implementaiton in `rowMaxs.IterableMatrix`

This will make it so our users can use `rowMaxs()` mostly without worry. (If they call `matrixStats::rowMaxs()` directly it won't work for BPCells objects, but any other copy of `rowMaxs` they call will support the full range of matrix types)

r/R/matrix.R
```diff
@@ -158,6 +157,36 @@ all_matrix_inputs <- function(x) {
...
+    res <- matrix_compute_max_per_row(iter)
+  }
+  names(res) <- rownames(x)
+  res
+}
+
+
+#' Get the max of each col in an interable matrix
+#' @param x IterableMatrix/dgCMatrix object
+#' @export
```
bnprks, line 177: One other nice trick we can do for the docs is using `@describeIn` or `@rdname` to group functions together into a single documentation page. For example, see [open_fragments_10x](https://bnprks.github.io/BPCells/reference/open_fragments_10x.html) docs and implementation.

For full docs see [roxygen2](https://roxygen2.r-lib.org/articles/reuse.html)

r/NEWS.md
```diff
@@ -79,6 +79,7 @@ well as all users who submitted github issues to help identify and fix bugs.
   graph-based clustering
 - Add `checksum()`, which allows for calculating an MD5 checksum of a matrix contents. Thanks to @brgrew for submitting this improvement in pull request #83
 - `write_insertion_bedgraph()` allows exporting pseudobulk insertion data to bedgraph format
+- Add `row_max()` and `col_max()` functions, which return the maximum value in each row or column of a matrix.
```
bnprks, line 82: This should go under the 0.2.1 section. (No guarantees we'll actually call that version 0.2.1, but we need some number there to get the documentation website to render properly)

r/R/matrix.R
immanuelazn: I think this makes interfacing with BPCells a lot more elegant, however for the future I wonder if doing this for each function will end up creating a lot of bloat.  Maybe we could create a function to automatically perform this strategy, given the default libraries with the same function names?

r/tests/testthat/test-matrix_stats.R
```diff
@@ -187,4 +187,67 @@ test_that("svds registers generic with RSpectra", {
...
+  i1_t <- t(m1) %>% as("dgCMatrix") %>% as("IterableMatrix")
+  i2_t <- m1 %>% as("dgCMatrix") %>% as("IterableMatrix") %>% t()
+  # for the case of pre-tranposed matrics prior to conversion to IterableMatrix
+  expect_equal(rowMaxs(i1_t), matrixStats::rowMaxs(as.matrix(t(m1))))
+  expect_equal(colMaxs(i1_t), matrixStats::colMaxs(as.matrix(t(m1))))
+  # case of transposing after conversion to IterableMatrix
+  expect_equal(rowMaxs(i2_t), matrixStats::rowMaxs(as.matrix(t(m1))))
+  expect_equal(colMaxs(i2_t), matrixStats::colMaxs(as.matrix(t(m1))))
+})
+
```
immanuelazn, line 236: The last test case here sort of encapsulates all cases together, by iterating through every test matrix, and transpose status.  Would you prefer the ones above, or the single one below?

bnprks: My personal preference is the single combined test since it's about 3x shorter and still pretty clear to read what's going on (though either approach seems perfectly fine)

r/R/matrix.R
bnprks: That is a good idea! I'm not quite sure about how to programmatically handle defining functions with a given set of arguments, but it's probably something we could figure out with the Advanced R book.

r/R/matrix.R
```diff
@@ -158,6 +157,76 @@ all_matrix_inputs <- function(x) {
...
+  iter <- iterate_matrix(convert_matrix_type(x, "double"))
+  if(x@transpose == TRUE) {
+    res <- matrix_max_per_col_cpp(iter)
+  } else {
+    res <- matrix_max_per_row_cpp(iter)
+  }
+  names(res) <- rownames(x)
+  return(res)
+}
+if (requireNamespace("MatrixGenerics", quietly=TRUE)) {
```
bnprks, line 191: This should be wrapped in `rlang::on_load` (so it runs at every package load, not just at install time; this handles installing BPCells then MatrixGenerics later)

r/NEWS.md
```diff
@@ -12,6 +12,7 @@ Contributions welcome :)
 - `apply_by_col()` and `apply_by_row()` allow providing custom R functions to compute per row/col summaries.
   In initial tests calculating row/col means using R functions is ~2x slower than the C++-based implementation but memory
   usage remains low.
+- Add `row_max()` and `colMaxs()` functions, which return the maximum value in each row or column of a matrix. If `matrixStats` or `MatrixGenerics` packages are installed, `BPCells::row_max()` will fall back to their implementations for non-BPCells objects.
```
bnprks, line 15: row_max -> rowMaxs

r/R/matrix.R
immanuelazn: Just realized I did that with one of the functions but not the other.. my bad!

r/R/matrix.R
```diff
@@ -158,6 +157,76 @@ all_matrix_inputs <- function(x) {
   x
 }
 
+#' Get the max of each row in an iterable matrix
+#' @param x IterableMatrix object/dgCMatrix object
+#' @return *vector of maxes for every row
+#' @describeIn IterableMatrix-methods Calculate rowMax (replacement for `matrixStats::rowMax()`)
```
bnprks, line 163: Just a couple small typos in these doc strings.
```suggestion
#' @return * `rowMaxs()`: vector of maxes for every row
#' @describeIn IterableMatrix-methods Calculate rowMaxs (replacement for `matrixStats::rowMaxs()`)
```
colMax also needs "colMax" -> "colMaxs" in @describeIn

r/tests/testthat/test-matrix_stats.R
immanuelazn: Cool, I'll keep that in mind when making test cases in the future.  Thanks!

PR #110, opened 2024-07-12
Add linear regression for `IterableMatrix`

ycli1995:
Hi, @bnprks,

I finally get `LinearResidual` works correctly. This idea was generally inspired from [ResidualMatrix](https://bioconductor.org/packages/3.19/bioc/html/ResidualMatrix.html) package.

A new test has been added to `test-matrix_transforms.R` to make sure `linear_regress_out` get the same results compared to `Seurat:::RegressOutMatrix`.


bnprks:
Hi @ycli1995, this looks really good, thanks! 

I'll probably have a few small suggestions once I've done a full review, likely around adding some detail to the documentation and thinking about the best function name and parameter names for R. I'll add those as comments via the github review interface as I come up with them.

I do have one additional suggestion which is optional, but would massively speed up running PCA on these matrices: if you also implement `vecMultiply(Right/Left)` and maybe even `denseMultiply(Right/Left)` on the `LinearResidual` class. I think `Shift.cpp` will be a useful reference here again for the useful Eigen functions to accomplish this. 

This will speed up PCA a bunch because explicitly calculating `X - Q * Qty` is pretty slow as it's a dense matrix. But if we calculate `(X - Q * Qty)v` as `(X * v) - Q * (Qty * v)` then we'll have a sparse multiply with X (fast), and two fast multiplies with the relatively-small `Q` and `Qty` matrices. I'd estimate at 10-20x speedups since RNA matrices are usually around 95% sparse.

r/R/transforms.R
```diff
@@ -812,3 +812,71 @@ multiply_cols <- function(mat, vec) {
...
+  Q <- as.matrix(qr.Q(qr))
+  col_params <- t(Q)
+  Qty <- t(as.matrix(mat %*% Q))  ## row_params = Qty
+  wrapMatrix(
+    "TransformLinearResidual", 
+    convert_matrix_type(mat, "double"),
+    row_params = Qty,
+    col_params = col_params,
+    global_params = numeric()
+  )
```
bnprks, line 864: You should add in a check for `storage_order(mat)` here, and if it's `"row"`, then I think you'll want to swap the row_params and col_params to make your code work out. (The C++ layer thinks of everything as colum-major, so an R matrix with `storage_order(mat) == row` will be seen by C++ as `t(mat)`. Luckily the math of this transform is nearly the same in C++ for `t(mat)`)

r/src/bpcells-cpp/matrixTransforms/LinearResidual.cpp
```diff
@@ -0,0 +1,33 @@
...
+    for (uint32_t i = 0; i < count; i++) {
+        double zero_val = 0;
+        // col_params = t(Q)
+        // row_params = Qty
+        // val = X - t(Qty) %*% t(Q)
+        // val = X - t(row_params) %*% col_params
+        Eigen::ArrayXXd prod_res = fit.row_params.col(start_row + i) * fit.col_params.col(col);
+        zero_val = -prod_res.sum();
+        values[i] = zero_val;
+    }
```
bnprks, line 30: This is very optional, but an interesting way to take advantage of the Eigen matrix operations. (honestly I wouldn't be surprised if it compiles to identical CPU instructions)
```suggestion
    Eigen::Map<Eigen::VectorXd> values_vec(values, count);
    // val = X - t(Qty) %*% t(Q)
    // val = X - t(row_params) %*% col_params
    values_vec = -fit.row_params.middleCols(start_row, count).matrix().transpose() *
                 fit.col_params.col(col).matrix();
```

bnprks (commented):
Here's some suggested changes I had going through a more detailed pass.

There are a couple naming changes I didn't put in as suggestions, but I would be happy to put in by adding my own commit to your branch if you're okay with them:
1. rename `linear_regress_out` to `regress_out` (aiming to keep the function name shorter and more similar to seurat / scanpy)
2. rename the `latent_data` argument to `features` (Since this function is not specific to single cell applications, I think a more general parameter name could be good)
3. Maybe add a `prediction_axis = "row"` default argument, which users could flip to "col" if the predicted variables (i.e. genes) were in the columns rather than rows. I'm not sure how much this argument is needed, though

Feel free to accept any suggestions you like, and let me know what you think of the proposed naming changes.

ycli1995:
Hi, @bnprks.

Thank you so much for those suggestions! I've push new commits for them.

About changing `latent_data` to `features`, currently the word "features" usually means "genes", "peaks" or something as row names of a single-cell matrix, especially in R (mainly in `Seurat`). Therefore, "features" may be not the right name for those confounding data in linear regression. In `Seurat`, confounding data to be regressed out are named as `latent data`, and in `scanpy` are named as `regressors`. I think keeping consistent with `Seurat` could be a good idea currently.

r/R/transforms.R
```diff
@@ -812,3 +812,84 @@ multiply_cols <- function(mat, vec) {
...
+  iterate_matrix_linear_residual_cpp(res, x@row_params, x@col_params)
+})
+setMethod("short_description", "TransformLinearResidual", function(x) {
+  # Return multiple lines, one for each transform active
+  res <- short_description(x@matrix)
+  if (any(nrow(x@row_params) == 0, nrow(x@col_params) == 0)) {
+    res <- c(res, sprintf("Linear regress out 0 variable"))
+    return(res)
+  }
+  res <- c(res, sprintf("Linear regress out %d variable(s)", nrow(x@row_params) - 1))
```
bnprks, line 838: This is a picky comment but I think will make the user experience a lot easier: could we make it so this will print out some of the actual variable names that are getting regressed out? I think this will require:

1. adding a slot to store the variable names
2.  initializing the variable names in `regress_out`
3.  printing it out here via a call to `pretty_print_vector(x@var_names)`

I don't think we can rely on `nrow(x@row_params)` because if someone passes in a factor to regress out, then `model.matrix` will expand it into indicator variables, but logically its just a single variable getting regressed out.

```suggestion
  res <- c(res, sprintf("Linear regress out %d variable(s)%s", length(x@var_names), pretty_print_vector(x@var_names, prefix=": ")))
```

r/R/transforms.R
```diff
@@ -812,3 +812,84 @@ multiply_cols <- function(mat, vec) {
...
+#' variable to regress out.
+#' @param prediction_axis Where the predicted values are. Options include "row" (default) and "col". 
+#' The row number of `latent_data` should be equal to number of the other axis than `prediction_axis` 
+#' in `mat`.
+#' @return An IterableMatrix. If input an empty `latent_data`, will return the `mat` itself.
+#' @export
+regress_out <- function(mat, latent_data = NULL, prediction_axis = c("row", "col")) {
+  if (length(latent_data) == 0) {
+    return(mat)
+  }
```
bnprks, line 857: I don't think we should silently do nothing in this case. Either we should take away the `latent_data = NULL` default and have the function error out with no input variables, or we should make the function regress out an intercept-only model (possibly with an additional warning printed) when 0 variables or NULL are passed in.

e.g. something like this (though it would have to come after the `match.arg(prediction_axis)` call and we'd have to add an option for `latent_data` to be a matrix as suggested below)
```suggestion
  if (length(latent_data) == 0 || is.null(latent_data)) {
    prediction_dim <- if(prediction_axis == "row") nrow(mat) else ncol(mat)
    latent_data <- matrix(rep.int(1, prediction_dim), dimnames=list(NULL, "(Intercept)"))
  }
```

bnprks (commented):
Hi @ycli1995, thanks for making those changes! 

I agree with your point that changing the parameter name `latent_data` to `features` might be confusing. Personally, I'd still prefer a different name like scanpy's `regressors` as I feel it is more standard terminology for linear regressions than `latent_data`. Though if you'd strongly prefer `latent_data` we can keep that name.

I've included a couple second-round (hopefully final) comments below. One thing outside of the inline comments is we'll need to add a few additional tests with the expanded functionality:

1. Add a test for the vector/matrix multiply ops (I'd recommend adding it to `test-matrix_utils.R` so you can use the `test_dense_multiply_ops` helper function, similar to the test "Dense matrix-vector multiply works" on line 461 but you don't need to test multiple inner_dim sizes; just take the default)
2. If you add the ability to pass a matrix instead of a data.frame, add a test confirming that works (just one basic case, doesn't need to re-cover all the variants tested with data.frame)
3. If you decide to allow an intercept-only regression (see comment below), then also confirm that works with one basic case.

Really great job on this overall, and I hope you don't mind me being very picky with my comments. I think after this round, if it's all right with you I can make one commit of my own crediting this in the `NEWS.md` file and tweaking the wording of the documentation. Then I think we should be basically ready to merge this into main!

ycli1995:
Hi, @bnprks

1. New commits have been added to ungroup the matrix multiplications and add corresponded R tests. Note that the results of multiplications passed `expect_equal` but not `expect_identical`, although I think `all.equal` is restrict enough. Therefore I added an option `test_func = expect_identical` for `test_dense_multiply_ops`.
2. I also agree that the variables to be regressed out should be printed to users. A new slot `vars_to_regress` has been added to `TransformLinearResidual`.
3. For `latent_data`, I think it's a good idea that we keep it as simple as possible currently:
* I agree that we should not silently do nothing when `latent_data` is empty. In other words, if one doesn't want to regress out anything, just don't need to call the function. Therefore I decided to keep `latent_data` as a compulsive parameter. 
* A model matrix and a bundle of variables (stored in a table) are totally different things. Treating both them as a single parameter `latent_data` may lead to unnecessary confusions. We just need to keep `latent_data` as what it is documented.
* Should we allow users to pass their own model matrix? In our current implementation, we mainly focus on regressing out the effects of provided variables with a linear model. That is what the calculations and usages for `col_params` and `row_params` are defined based on, as well as how the `model_mat` is constructed from `latent_data`. I've tried to make it more clear in the documents.
* In the future, if we want to add supports to other methods, the implementions can be carried by independent functions.

Thanks again for those great suggestions! I think after we add the linear regression feature to `BPCells`, we can finally make almost all the general data pre-processing for large scale scRNA-seq on-disk.

bnprks:
Hi @ycli1995, thank you for those additional changes. For (1), using `expect_equal` is fine here. For (2) looks good. For (3) your choice for a simple solution seems good for now. If people want to pass their own model matrix in the future it will be easy to update to allow that.

I double-checked the whole test suite passes on my end, and made a couple additional changes reflected in my commits above:

- Add explicit parentheses back to LinearResidual (just the opposite grouping to how they were originally). Without the parentheses still half of the ops would have the wrong grouping due to C++ multiplication operator associativity.
- Edit function documentation, focusing on adding a "details" section
- Tweak the `short_description` message
- Slightly reduce the test data dimension since I think they are already more than enough to catch errors.
- Add news and update the website docs

Let me know if this all looks okay to you, in which case I'll merge this in!

ycli1995:
Hi, @bnprks 

Thanks for the double-checks. Those improvements look great! 

For your concerns about mean and variance calculations, I think users can take advantages of setting `threads` in `matrix_stat()` to speed them up. In fact, I roughly tested `regress_out` on the 10x Genomics dataset of 1.3 million mouse brain cells with 2000 highly variable genes, and it just took several minutes to run `matrix_stats` with `thread = 8L`. And the memory usage was still keep within 10GB, what a wonderful package `BPCells` is!

bnprks:
Great, fully merged in now and the docs website is updated now as well. Thanks for your work on this!

PR #114, opened 2024-07-25
[r] Fix `trackplot_coverage()` when interacting with single clusters

immanuelazn:
# Description
Changes `trackplot_coverage()` so to prevent data conversion from tibble to named vector when dealing with only one cluster.

# Tests
- Added test cases for `trackplot_coverage()` to check for crashing out in both multi-cluster and single-cluster cases
- Added test for `trackplot_coverage()` return data type


bnprks:
Looks great! Thanks for adding some extra test coverage + fixing some NEWS typos along the way

PR #115, opened 2024-07-30
[r] fix trackplot bugs with small tracks/motifs

immanuelazn:
# Description
Various small changes to coverage plots and genome peak plots to handle smaller motifs better.

- Created at least a single arrow for peaks that are not long enough for the pre-determined minimum arrow segment length.  Note, in very small peaks the arrow may go out of bounds of the peak itself, which will likely also need to be fixed in the future.
- Fixed an issue where the head_only boolean flow in `trackplot_create_arrow_segs()` did not properly assign values to the start and end position if a strand was negative.
- Fixed an issue arising from tracks smaller than 500 bp in `trackplot_coverage()`

# Tests
- Added a small test case for the head_only boolean flow in `trackplot_create_arrow_segs()`
- Added tests for small regions in `trackplot_coverage()`

### DIFF
--- a/r/R/trackplots.R
+++ b/r/R/trackplots.R
@@ -188,14 +188,11 @@ trackplot_create_arrow_segs <- function(data, region, size = 50, head_only = FAL
   if (head_only) {
     # Set segment size small enough to be invisible if head_only
     # Cannot swap values with if-else mutation block due to sequential evaluation
-    arrows_shortened <- dplyr::mutate(
-      arrows,
-      start = ifelse(strand, arrows$end - 1e-4, arrows$start),
-      end = ifelse(strand, arrows$end, arrows$start+1e-4)
-    )
     arrows <- arrows %>% dplyr::mutate(
-      start = arrows_shortened$start,
-      end = arrows_shortened$end
+        start_tmp = start,
+        start = ifelse(strand, end-1e-4, start),
+        end = ifelse(strand, end, start_tmp+1e-4),
+        start_tmp = NULL
     )
   }
   arrows <- dplyr::filter(arrows, start >= region$start, start < region$end, end >= region$start, end < region$end)
@@ -390,7 +387,7 @@ trackplot_coverage <- function(fragments, region, groups,
   }
   colors <- colors[seq_len(ncol(membership_matrix))]
 
-  bin_centers <- seq(region$start, region$end - 1, region$tile_width) + region$tile_width %/% 2
+  bin_centers <- seq(region$start, region$end - 1, region$tile_width) + ((region$tile_width-1) / 2)
   bin_centers <- pmin(bin_centers, region$end - 1)
 
   mat <- (tile_matrix(fragments, region, explicit_tile_names = TRUE) %*% membership_matrix) %>%
@@ -413,7 +410,7 @@ trackplot_coverage <- function(fragments, region, groups,
 
   ymax <- quantile(data$normalized_insertions, clip_quantile)
   # Set precision of y-axis range label to within 1% of the max value
-  ymax_accuracy <- 10^as.integer(log10(0.01 * ymax))
+  ymax_accuracy <- 10^floor(log10(0.01 * ymax))
   range_label <- sprintf("[0-%s]", scales::label_comma(accuracy = ymax_accuracy, big.mark=" ")(ymax))
 
   data$normalized_insertions <- pmin(data$normalized_insertions, ymax)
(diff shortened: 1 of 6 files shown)


immanuelazn:
For peaks with no arrows:
Before
image
After
image

r/R/trackplots.R
```diff
@@ -180,14 +180,22 @@ trackplot_create_arrow_segs <- function(data, region, size = 50, head_only = FAL
...
+      arrows,
+      start = ifelse(strand, arrows$end - 1e-4, arrows$start),
+      end = ifelse(strand, arrows$end, arrows$start+1e-4)
+    )
     arrows <- arrows %>% dplyr::mutate(
-      start = ifelse(strand, end - 1e-4, start),
-      end = ifelse(strand, end, start + 1e-4)
+      start = arrows_shortened$start,
+      end = arrows_shortened$end
     )
```
bnprks, line 199: On a third look at this, there is no bug in the original -- the only time the `end = ` case reads the `start` variable is if `strand` is `FALSE`, but in that case `start` hasn't been modified. I discovered this when trying to add tests for the `head_only` case and confirming the previous code was actually broken.

That said, for an approach that would be less dependent on the specific logic in this case I found a weird-looking alternative. It is still not great but a tiny bit simpler:

```r
arrows <- arrows %>% dplyr::mutate(
    start_tmp = start,
    start = ifelse(strand, end-1e-4, start),
    end = ifelse(strand, end, start_tmp+1e-4),
    start_tmp = NULL
)
```

Confirmed to work on a test tibble `arrows <- tibble(start=1:3, end=11:13, strand=c(TRUE, FALSE, TRUE))`

r/R/trackplots.R
```diff
@@ -382,7 +390,7 @@ trackplot_coverage <- function(fragments, region, groups,
   }
   colors <- colors[seq_len(ncol(membership_matrix))]
 
-  bin_centers <- seq(region$start, region$end - 1, region$tile_width) + region$tile_width / 2
+  bin_centers <- seq(region$start, region$end - 1, region$tile_width) + region$tile_width %/% 2
```
bnprks, line 393: Looking back at this, I think the original calculation was at least off by 0.5 (i.e. a width-1 bin shouldn't be shifted, and a width-2 bin should be ideally shifted by 0.5bp). Based on that I'd propose `(region$tile_width-1) / 2`.

Though if you think the off-integer placement looks bad keeping the `%/%` is a good way to solve that.

r/tests/testthat/test-trackplots.R
```diff
@@ -180,44 +192,54 @@ test_that("trackplot_coverage doesn't crash", {
...
-              groups=frags_metadata$cell_type,
-              cell_read_counts=frags_metadata$atac_reads,
-              bins=500
-            )
-        )
+    # for small motifs
+    region2 <- tibble::tibble(
+        chr = "chr15",
+        start = 31283779,# 31277136,
+        end = 31283870#31322748
```
bnprks, line 205: Shouldn't have commented-out code left in the tests

r/tests/testthat/test-trackplots.R
```diff
@@ -180,44 +192,54 @@ test_that("trackplot_coverage doesn't crash", {
...
+        expect_no_condition(
+            ggplot2::ggplot_build(trackplot_coverage(
+                  frags_one_cluster,
+                  region=region,
+                  groups=frags_metadata_one_cluster$cell_type,
+                  cell_read_counts=frags_metadata_one_cluster$atac_reads,
+                  bins=500
+                )
+            )
+        )
```
bnprks, line 231: Running this test on `region2`, I get the following error:
```
Error: Expected `ggplot2::ggplot_build(...)` to run without any conditions.
ℹ Actually got a <simpleWarning> with text:
  NAs introduced by coercion to integer range
```
Can you track down why this is?

bnprks (commented):
Overall looks good! Thanks for the before/after screenshots. A couple comments below.

I have two other changes I'd like to add in via a commit:
1. Add test cases for `trackplot_create_arrow_segs(head_only=TRUE)`. I wrote these while thinking through the `dplyr::mutate` logic, so might as well toss them in. (I don't like that they are dependent on our specific choice of 1e-4, but couldn't think of another option)
2.  I ran a global search for `install_github("bnprks/BPCells")` and found two more spots to update the text

r/R/trackplots.R
immanuelazn: It was moreso in the case of having tile width of 1, where region < 500.  I looked at the vector and it would always be duplicates with additions of 1 (ie c(10, 10, 12, 12, 14, 14)). Therefore we lost the bin centers for half of the vectors.  Your solution overall works better I think though, but I would make sure that bin_centers isn't being outputted as an integer list.

r/tests/testthat/test-trackplots.R
immanuelazn: Oops!

PR #117, opened 2024-07-31
[r] Allow logical mask in select_cells/chromosomes

bnprks:
Pretty simple usability tweak here, just making it possible to use a TRUE/FALSE vector to select cells or chromosomes


immanuelazn (approved):
Short and sweet, good work!

PR #118, opened 2024-08-13
Macs peak calling

immanuelazn:
# Description
Add macs2/macs3 peak calling
Utilize a multi-threaded approach separated by cluster within `prep_macs_inputs()`.  Allow users to choose a specific step to run (ie "prep-inputs", "run-macs", "read-outputs") to allow for executing macs calls in external slurm cluster.  
Also change behaviour in `InsertionIterator` to not keep insertions in memory after moving to next fragment.

# Tests
- Added test cases for prep_macs() parameterized by cluster existence, threading, and insertion_mode


bnprks:
Here's some initial comments I came up with from a read-through. Overall looks like a pretty good structure and approach. I'm focusing on some higher-level comments here. 

There are also several small things that I think could be tweaked around choices of naming, docs, default arguments, organization etc. Rather than dictate solutions to those while you're still drafting, I'll let you finish up and do your own polishing pass before we discuss those.

C++ comments:

- I'm pretty sure that `writeInsertionBedByPseudobulk` should write out duplicate coordinates rather than explicitly skipping printing repeats
- Once we get a chance to do a performance profile, we should check if `gzprintf` is taking up a bunch of time (and comparing with/without a `.gz` file ending to isolate if the compression or the `printf` part is slow)
  - We might also want to take a quick look to see if `macs3` can support any other kinds of compressed input that would be faster to write
 - We will definitely need to debug the memory issue you identified in `InsertionIterator` as well

R comments

- With the option to split up execution using `step`, I'm a bit concerned about errors that might happen if users edit the input folder between steps, or try calling a later step with arguments that are mismatched/incompatible with their earlier step
  - This might mean making `cell_names` non-nullable, and maybe even `fragments`, though that might not be worth it given the usability trade-off.
- `prep-inputs` should probably write the macs command to a `.sh` file (one-per-cluster). Also returning the commands is as a vector is fine, but later steps could then read from the `.sh` files to simplify 
  - Also we usually have a few more arguments present, including e.g. `--shift`. You might want to check the ArchR source a bit more for that as I think arguments are listed in at least two separate places 
- For `mclapply` we probably also want to always use the argument `mc.preschedule=FALSE`

immanuelazn:
There are a few more tests to be added that directly compare writing insertions from fragments in ArchR vs BPCells.  I might just copy the output files to eliminate the dependence of building an entire ArchR project.  
Otherwise, I think I've addressed most of your comments, and the feature is in a relatively polished state.

bnprks:
Looking pretty good! Most remaining comments are style-related, though I did end up with a large number of those in `call_macs_peaks()`. Trying out a bulleted list here in rough order of the code rather than using Github's inline comment feature since it's a bit easier for me to prep.

Good job setting up the priority queue `InsertionIterator` -- I'm fine leaving it as a stopgap for this PR but I'd really like to get the radix sort algorithm fixed up and running soon after we merge this.

`InsertionIterator`

- Very cool that you were able to compare performance between the priority queue and the sorting version!
- Minor style note but `*(starts + i)` is the same as `starts[i]` 

`writeInsertionBedByPseudobulk()`

- Given that this writes out just one pseudobulk, I'm not sure why both the `cells` and `group_of_interest` arguments are needed. Arguably, you could even just pass in `fragments` to already be subset to just the cells of interest from the R side with `select_cells()`
  - At least filtering prior to `InsertionIterator` will save some sorting work
- Unclear if `keep_dups` argument is needed (more discussion below). This also might eliminate the need for the `cleanup at end of chromosome` section
- Error messages shouldn't be prefixed with `writeBedgraph:` anymore
-  I'd suggest just calling this function just `writeInsertionBed` for simplicity. As a cleanup task I can edit `writeInsertionBedgraph` to just write one pseudobulk at a time anyhow as I think that will make the most sense
- Either this file name will need an update or this function should go into its own file because it doesn't handle `Bedgraph` files. (I'd be happy to rename to maybe `BedWriter` since bedgraph is kinda just another type of bed file)

`prep_macs_inputs()`

- Does this need to be a separate exported function? Seems like it could mostly slot in to `call_macs_peaks` or at least be private



`call_macs_peaks()`

1. Make docs and parameter names consistent with `call_peaks_tile` (cell_groups and effective_genome_size arguments). Can use `@inheritParams` to save on re-typing docs if the text will be identical.
   - You might also consider copying the default initialization of `cell_groups`, though I'm not sure how that would work if we allow `fragments` to be NULL
2. `path` docs. Put directory structure in `@details` section. Specify inputs are `<path>/input/<cluster>.bed(.gz)` and shell commands are `<path>/input/<cluster>.sh`
3. `insertion_mode` docs don't quite read naturally. If you want consistency with `write_insertion_bedgraph` I'd update both to say maybe "for coverage calculation" rather than "for insertion counts calculation"
4. Do we need the `keep_dups` option? I was just looking through the ArchR code and I don't think it would ever de-duplicate (and as mentioned previously I think it's likely undesirable to remove duplicate insertion sites in a peak calling context). Though if your testing indicates otherwise we can discuss
5. `step` argument should default to `all`
6. `macs_version` could probably be `macs_executable` instead, and allow the user to pass a full path if desired. (And I'd say default to macs3 at this point, or make the default NULL and try to auto-detect macs2 vs macs3 with test `system()` calls)
   - Either way, we probably want to check if macs can be run before prepping any inputs in the case of `step = all`  
7. `additionalParams` should be snake case like you have it in the docs. It also might be nice if the name made it clear that these are parameters getting passed to MACS
   - Still don't have a `--shift` argument being passed, which might make sense to add here
8. `quiet`: If we multi-thread `run-macs`, then we might want to write the macs stdout/stderr to a log file in `outputs/` and have this just print when we start writing each input file and when we start running each copy of macs (maybe also printing when each finishes?). Also, for consistency this should flip and be named `verbose`
9. `use_gz` docs would be helpful if they could give some absolute file sizes + time to write with /without compression for a small-ish example dataset (e.g. one of the public 10x pbmc datasets)
10. `threads` argument in `call_macs_peaks` should probably also be used in the `run-macs` section
11. Return value should be a single data frame, with an extra column specifying the group that the cluster was called in. Try to place the group column early in the ordering (probably either first or right before name?). A list of data frames is a little awkward to deal with as output
12. Will the `system2` call work properly when there are spaces in cluster names? Might want to add some quoting to `macs_call_template`. Also maybe get rid of any `/` character in a cluster names?
13. In docs, `bedfile` should be split up as two words for style

`tests`

- Looks reasonable to me. I like your tests that the insertion bed file has the right contents. If `prep_macs_inputs` stops being a public/separate function, you could adjust to just call `call_macs_peaks(step="prep-inputs")` in that test

immanuelazn:
I think I'm happy with the current state, but please let me know if you see any more glaring changes I should be making.  It will be a little bit of a journey attempting to extract insertion beds from ArchR, while removing the hardwired pseudobulk replicates and tn5 bias steps.  To make sure I can continue progressing, I suggest that I table creating a testcase from comparing outputs from ArchR because getting ArchR to skip those steps will require a decent amount of software engineering time by itself.  Would that be fine?

immanuelazn:
Ok it should be ready! I added in the test comparing Archr and BPCells results.  The only change required is setting `devtools::load_all()` to target your bpcells/archr install.

bnprks:
Overall, I like how this is going and fundamentally it's quite close to being ready. A lot of the details I asked for from the last round are fixed. I like the ordering you chose for the `call_macs_peaks` parameters, and making `write_insertion_bed` have a matching interface to `write_insertion_bedgraph`. The performance testing and comparison against ArchR outputs I imagine were tricky and time-consuming, but good job on them.

There are several more detail-oriented issues remaining though, several from my last round of suggestions which were either forgotten or only partially addressed.  I've also included a few new ones that came up as I was able to try out the function more.

Sorry for having taken so much of the week to get back to you on this, I hope it hasn't been too much of a blocker. I'd be happy to do a call early next week to discuss any questions/clarification about this round of comments.

<details>
<summary>Follow-ups still to be addressed from last comment round</summary>

`writeInsertionBed`

- After removing the `keep_dups` argument, there's no reason to have the variable `last_base` and hence the end-of-chromosome cleanup step. 

`call_macs_peaks()`

- `path` docs in details section. 
  - For input files, I'd describe contents/purpose of the files. E.g just "shell commands" doesn't say what they are for. "Bed files" doesn't explain what the contents are. 
  - For outputs, for the files BPCells doesn't use, we can just link to the macs3 documentation on output files. 
  - Very useful to say we only read the `narrowPeaks` file, but you need an extra line break so it doesn't get swallowed into the last bullet when rendered as a webpage
  - In the docs for the `path` argument, you should add a (see Details for directory structure) or equivalent message
- `macs_version` was not renamed in the function declaration, only in the docs. Also, using `match.arg` means that it's not possible for the user to pass a custom path for a macs executable
- `use_gz` docs would be helpful to give numbers for the time trade-off as well. Most straightforward would be to give both time + input folder size for with and without compression.
- `threads` argument in `call_macs_peaks` should probably also be used in the `run-macs` section
- Return value "cluster" column would ideally not be the last column
- Will the `system2` call work properly when there are spaces in cluster names? Might want to add some quoting to `macs_call_template`. Also maybe get rid of any `/` character in a cluster names?

</details>

New comments:

<details>
<summary>Verbose logging setup:</summary>

- `log_message` variable will not be defined in "run-macs" section if run standalone (and I also don't think it's necessary to include that information in the log file)

- In `write_insertion_bed()`, the logging should be inside the `mclapply` call so it prints at the correct time. (Despite the theoretical race condition, I think this basically always works out fine even in very tight loops)

  - The style of log you use in `run-macs` is better also, rather than printing out cluster names on individual lines

  - Example of the current (sub-optimal) output

    ```
    [1] "2024-08-23 16:10:15.508477 Writing bed file for clusters:"
    [1] "consonant"
    [1] "vowel"
    [1] "2024-08-23 16:10:15.532154 Finished writing bed files"
    ```

- I'd suggest `format(Sys.time())` to avoid having the fractional seconds get printed

- I think using `message()` rather than `print()` will be better to avoid the `[1]` prefixes getting printed

</details>

<details>
<summary>run-macs and error checking:</summary>

- `macs_executable` now has a slightly weird combination of both auto-detect and a default value. 

  I'd say the best option is to default NULL, and in the case of NULL we can do your auto-detect procedure. If not null, we just check that `macs` can be found else we error out

  - Also, the existing `stop_run` and `log_message` variables can probably be eliminated since I don't think there's a need / easy way to hang on to `log_message` for use in later steps.

- The existing check on `macs_executable` should check the output of `macs3 --version` to make sure it contains the text `macs` -- otherwise the check will pass for almost any executable even if it's not macs

- In `run-macs`, it's probably better for `system2` to write stdout/stderr directly to the log file so that logs are written as the command is running rather than all at the end. It's OK if this prevents writing the command as it appears MACS already prints the input command at the top of the log file

- Would be good to check in `run-macs` that the command ran without error (had a 0 exit code)

</details>

<details>
<summary>Other:</summary>

- The test for `macs_e2e_works` now introduces a system dependency on having `macs2` installed. Can we make it skip instead if `macs` is not installed?
- `--call-summits` probably can go in `--additional-params`, as it seems to just affect the peak calling method used by macs while leaving the output file format consistent
- in `write_insertion_bed`, `cell_groups` should probably come after `path` for consistency with `write_insertion_bedgraph`, and `cell_groups` should maybe default to NULL for consistency.

</details>
<details>
<summary>Questions/discussion areas (can do via call/slack)</summary>

- Is there a better name possible for `use_gz`? I'm just realizing the name doesn't really give any hints about what is getting compressed, so it would be nice to have a name that would be more specific to the macs bed insertion inputs. Let me know if you have a suggestion here 
- What do you think of the adjusting column names/values in the output? Reading the [[UCSC narrowPeak format docs](https://genome.ucsc.edu/FAQ/FAQformat.html#format12)](https://genome.ucsc.edu/FAQ/FAQformat.html#format12), I think it's kinda weird that `pValue` is given as -log10(p), and peak/pointSource is given relative to `start`. On the one hand it's nice to maintain a direct correspondence with the macs output, but some parts of it are kind of non-intuitive
</details>

r/R/atac_utils.R
```diff
@@ -593,6 +603,8 @@ call_macs_peaks <- function(fragments, cell_groups, path, effective_genome_size
                                                       "pValue", "qValue", "pointSource"),
                                           show_col_types = FALSE)
       peaks[[cluster]]$cluster <- cluster
```
bnprks, line 605: Could we actually name this column "group"? I'm not 100% sure on this, but I'm trying to think of consistency with the `cell_groups` argument name and not assuming that the cell groups correspond directly to clusters. 

I should have mentioned this on the previous review round, sorry

r/tests/testthat/test-atac_utils.R
```diff
@@ -248,6 +248,11 @@ test_that("write_insertion_bed works", {
 })
 
 test_that("macs_e2e_works", {
+  # only run if macs2 is installed
+  if ((suppressWarnings((system2("macs2", args = "--version", stdout = FALSE, stderr = FALSE) != 0)) ||
+      (!grepl("macs", system2("macs2", args = "--version", stdout = TRUE))))) {
+    skip("macs2 not installed")
+  }
```
bnprks, line 255: Two requested changes in this test:
1. Can we make it work with either macs2 or macs3 and just skip if neither is available?
2. Could we make the `cell_groups` in this test be `v o w e l` and `consonant;` to help check that our quoting is working properly?

r/R/atac_utils.R
```diff
@@ -459,61 +458,63 @@ write_insertion_bed <- function(fragments, cell_groups,
...
   insertion_mode <- match.arg(insertion_mode)
   step <- match.arg(step)
-  macs_version <- match.arg(macs_version)
   cell_groups <- as.factor(cell_groups)
-  stop_run <- FALSE
 
   # Change cell group spaces into underscores
   levels(cell_groups) <- gsub(" ", "_", levels(cell_groups))
-  if (use_gz) {
+  levels(cell_groups) <- gsub("/", "_", levels(cell_groups))
```
bnprks, line 516: I don't think escaping spaces should be needed if our quoting is working properly. I'll write up a little helper for checking for illegal characters like `/` and handling that properly

bnprks (commented):
Good set of changes! A couple small fixes are still needed as marked below. Then I have 2 discussion points for maybe some final polish

1. After playing around some, I think we should probably write the shell files for steps `prep-inputs`, `all`, and `run-macs`. This will ensure that if you change the macs parameters when calling `run-macs` those will be respected
2. I don't like how the existing escaping system will silently change group names in a way that produces changes in the output group/cluster names column. Windows has some very strange edge cases here, so I might take a stab at making a path normalization helper function that will spit out a warning for cases when we have to change the name

I'm still really tempted to adjust the column names in our output, even if we leave the values the same as you prefer. My ideal column name changes would be:
-  signalValue -> fold_enrichment
- pValue -> log10_pvalue
- qValue -> log10_qvalue
- pointSource -> summit_offset

I think something along those lines would give clearer names while avoiding having to alter the values

r/R/utils.R
```diff
@@ -44,3 +44,9 @@ document_granges <- function(
     "%s"
   ), intro_noun, bullets)
 }
+
+# Function which prints a message using shell echo
+# Useful for printing messages from inside mclapply when running in Rstudio
+message_parallel <- function(msg){
+  system(sprintf('echo "%s"', paste0(msg, collapse="")))
+}
```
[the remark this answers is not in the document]
immanuelazn: I think adding flexibility would be a good middle ground so we can utilize messages without timestamps through an argument

r/R/utils.R
```diff
@@ -45,8 +45,11 @@ document_granges <- function(
...
-# Function which prints a message using shell echo
-# Useful for printing messages from inside mclapply when running in Rstudio
-message_parallel <- function(msg){
+# Function which prints a message using shell echo.
+# Useful for printing messages from inside mclapply when running in Rstudio.
+log_progress <- function(msg, add_timestamp = TRUE){
+  if (add_timestamp) {
+    msg <- paste0(format(Sys.time(), "%Y-%m-%d %H:%M:%S "), msg)
+  }
   system(sprintf('echo "%s"', paste0(msg, collapse="")))
```
bnprks, line 54: I just did a bit of extra research -- looks like it is possible to explicitly detect running in RStudio via `.Platform$GUI == "RStudio"`. This might allow use `message()` to play a bit more nicely with R's normal systems while just resorting to `system('echo')` on RStudio

bnprks (commented):
Down to I think just two small tweaks needed left!

I also have some edits to the docs I'm considering, which would cover:
- Adjusting how we communicate the type and defaults for parameters
- Changing language from "cluster" to "cell group" in parameter docs
- Adjusting the intro language and making a section on running MACS manually

I can maybe show you the diff on a call rather than just pushing it here, since it's not 100% obvious this is how to do it


Then, I have two ideas on parameter defaults:
- `cell_groups = rlang::rep_along(cellNames(fragments), "all")` (for consistency with the `cell_groups` argument in other functions
- Either move `path` to be the second argument, or set it to default `path = tempfile("macs-peaks")`. The default is a bit risky in case someone wants to load their peaks back at a later point in time

r/R/atac_utils.R
```diff
@@ -615,6 +590,33 @@ call_macs_peaks <- function(fragments, cell_groups, path, effective_genome_size
   }
 }
 
+#' Test if MACS executable is valid.
+#' If macs_executable is NULL, this function will try to auto-detect MACS from PATH, with preference for MACS3 over MACS2.
+#' If macs_executable is provided, this function will check if MACS can be called.
+#' @return MACS executable path.
+#' @inheritParams call_macs_peaks
+#' @keywords internal
+macs_path_is_valid <- function(macs_executable) {
```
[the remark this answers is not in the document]
immanuelazn: Yeah I think that would've cleaned up the code too.  This was pretty easily modularizable and was taking up a decent bit of logic anyways so I think either (or both) ideas work

r/R/utils.R
immanuelazn: Interesting, let me put in some better handling for this

r/R/atac_utils.R
```diff
@@ -431,21 +431,21 @@ write_insertion_bed <- function(fragments, path,
   } else {
     assert_len(cell_groups, length(cellNames(fragments)))
     cell_groups <- as.factor(cell_groups)
+    levels(cell_groups) <- normalize_unique_file_names(levels(cell_groups))
```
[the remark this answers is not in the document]
immanuelazn: I was moreso thinking on the side of having this still function independently.  ie for some reason `write_insertion_bed()` was being used separately from `call_macs_peaks()` it will still be able to work with haphazardly formed cell_group names.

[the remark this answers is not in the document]
bnprks: I think the code as is actually breaks users with weird `cell_group` names rather than fixing them -- `path` is already supposed to hold valid path names, and the responsibility for that is very clearly on the user. But if we mess with `levels(cell_groups)`, then it might not line up anymore with the names of `path`.

r/R/atac_utils.R
```diff
@@ -379,30 +379,260 @@ call_peaks_tile <- function(fragments, chromosome_sizes, cell_groups = rep.int("
...
+  insertion_mode <- match.arg(insertion_mode)
+  step <- match.arg(step)
+  cell_groups <- as.factor(cell_groups)
+  levels(cell_groups) <- normalize_unique_file_names(levels(cell_groups))
+  path <- normalizePath(path)
+  path_bed_input <- paste0(path, "/input/", levels(cell_groups), ".bed.gz")
+  names(path_bed_input) <- levels(cell_groups)
+  path_macs_output <- paste0(path, "/output/", levels(cell_groups))
+  # Create paths
+  dir.create(file.path(path, "input"), showWarnings = FALSE, recursive = TRUE)
```
bnprks, line 535:
```suggestion
  # Create paths
  dir.create(file.path(path, "input"), showWarnings = FALSE, recursive = TRUE)
  path <- normalizePath(path)
  path_bed_input <- paste0(path, "/input/", levels(cell_groups), ".bed.gz")
  names(path_bed_input) <- levels(cell_groups)
  path_macs_output <- paste0(path, "/output/", levels(cell_groups))
```
I think this slight re-order is better, thinking about the case where `path` is a not-yet-existing directory. `dir.create` will do the `~` expansion, but `normalizePath()` will error out on a non-existent path (or else operate with reduced efficiency if `mustWork=FALSE`)

immanuelazn: Agreed, I was getting a ton of warnings on `normalizePath()` on the bed section so this would make sense on `call_macs_peaks()`

r/R/atac_utils.R
```diff
@@ -379,30 +379,260 @@ call_peaks_tile <- function(fragments, chromosome_sizes, cell_groups = rep.int("
...
     path <- path[levels(cell_groups)]
     cell_groups <- as.integer(cell_groups) - 1L
   }
   write_insertion_bedgraph_cpp(iterate_fragments(fragments), cell_groups, path, insertion_mode)
 }
 
+#' Create bed files from fragments split by cell group.
+#' @param path (character vector) Path to save bed files. If `cell_groups` is provided, this must be a character vector with one name for each level in `cell_groups` 
+#' Else, this must be a character vector of length 1.
+#' @param cell_groups (character vector or factor) Cluster assignments for each cell. If is null, all cells are treated as one group.
```
bnprks, line 412:
```suggestion
#' @param cell_groups (character vector or factor) Cluster assignments for each cell.
```
Looks like we no longer have the `NULL` option so this can come out of the docs

r/NEWS.md
```diff
@@ -18,6 +18,7 @@ Contributions welcome :)
 - Add `regress_out()` to allow removing unwanted sources of variation via least squares linear regression models.
   Thanks to @ycli1995 for pull request #110
 - Add `trackplot_genome_annotation()` for plotting peaks, with options for directional arrows, colors, labels, and peak widths. (pull request #113)
+- Add MACS2/3 input creation and peak calling through `prep_macs_inputs()` and `call_macs_peaks()`
```
bnprks, line 21:
```suggestion
- Add MACS2/3 input creation and peak calling through `call_macs_peaks()`
```

r/pkgdown/_pkgdown.yml
```diff
@@ -74,6 +74,7 @@ reference:
   - call_peaks_tile
   - merge_peaks_iterative
   - write_insertion_bedgraph
+  - call_macs_peaks
```
bnprks, line 77:
```suggestion
  - call_macs_peaks
  - call_peaks_tile
  - merge_peaks_iterative
  - write_insertion_bedgraph
```
Re-order `call_macs_peaks` to slot in next to the other peak-calling oriented function (but coming first since it's a lot better)

r/src/bpcells-cpp/fragmentUtils/BedWriter.cpp
```diff
@@ -6,17 +6,53 @@
...
-
 namespace BPCells {
 
+// Write bedgraph files including chr, start, end with only 1bp insertions,
+// only considering a subset of cells given by the cells vector. Leave duplicates in the bedfile.
+// Args:
+// - fragments: source of fragments to convert to insertions
+// - output_path: The file path to save the bedgraph
+// - mode: StartOnly = include only start coords, EndOnly = include only end coords, Both = include start + end coords
+// - keep_dups: If true, keep duplicate insertions in the bedfile
```
bnprks, line 23:
```suggestion
```

r/src/bpcells-cpp/fragmentUtils/BedWriter.h
```diff
@@ -22,6 +22,21 @@ enum class BedgraphInsertionMode {
...
 };
 
+
+// Write bedgraph files including chr, start, end with only 1bp insertions,
+// only considering a subset of cells given by the cells vector. Leave duplicates in the bedfile.
+// Args:
+// - fragments: source of fragments to convert to insertions
+// - output_path: The file path to save the bedgraph
+// - mode: StartOnly = include only start coords, EndOnly = include only end coords, Both = include start + end coords
+// - keep_dups: If true, keep duplicate insertions in the bedfile
```
bnprks, line 32:
```suggestion
```

r/tests/testthat/test-atac_utils.R
```diff
@@ -154,4 +151,186 @@ test_that("write_insertion_bedgraph works", {
...
+    verbose = FALSE,
+    threads = 2
+  )
+  # Check to see if the output files are created
+  for (cluster in c("v o w e l", "consonant;")) {
+    expect_true(file.exists(file.path(dir, "output", cluster, paste0(cluster, "_peaks.narrowPeak"))))
+    expect_true(file.exists(file.path(dir, "output", cluster, paste0(cluster, "_peaks.xls"))))
+    expect_true(file.exists(file.path(dir, "output", cluster, paste0(cluster, "_summits.bed"))))
+  }
+  # Read i%sn the outputs
```
bnprks, line 309:
```suggestion
  # Read in the outputs
```

PR #120, opened 2024-08-15
[r] Add randomized test to detect dimnames errors

bnprks:
I thought I had found another dimnames bug recently, but it turns out it was already fixed in the current main branch. In the mean time, I had already written some randomized tests that will hopefully provide some extra test coverage against future dimnames regressions. 

It is a little slow to run, so I've limited it to a number of randomized tests where there's an 80% detection rate on the old bug, which should take approximately 0.1 seconds.

### DIFF
--- a/r/tests/testthat/test-matrix_utils.R
+++ b/r/tests/testthat/test-matrix_utils.R
@@ -219,6 +219,61 @@ test_that("Subset rbind/cbind dimnames are preserved (#100 regression)", {
   expect_identical(m1[1:5,1:5], as(write_matrix_memory(m2), "dgCMatrix"))
 })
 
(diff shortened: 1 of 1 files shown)


immanuelazn:
The logic looks intuitive and elegant.  However I worry after running testthat numerous times that sometimes the debugging messages are not that useful. i.e. no information of which args were passed to a function etc.    In this case, I think that may cause some pain when trying to isolate which specific operations are giving you issues, and randomized testing might further exacerbate this.

Understandably it is not hard to add in a few print statements if the tests _do_ end up going wrong, but just sharing my two cents

immanuelazn (approved):

PR #123, opened 2024-08-29
[r] Fix Rcpp warning about out-of-bounds access

bnprks:
Rcpp started warning about out-of-bounds access stating `subscript out of bounds (index 0 >= vector size 0)` on some tests. This was due to length-zero arrays in BPCells matrices arising from e.g. single element arrays with perfectly compressed data arrays (0 bits per element in the bitpacking data array).

Now we avoid accessing an index of a possibly-empty array and use an alternate method to get a data pointer.


immanuelazn (approved):
Looks good to me!

Issue #86, opened 2024-05-01
Error of hdf5 when installing on HPC Rstudioserver


yifanfu01:
Hi,
Many thanks for this useful package.

I use HPC without Root access, and there used **module** control to set system environment, before enter the R environtment, I load:
```

apps/hdf5/1.8.13
compiler/gcc/9.1.0
mathlib/hdf5/1.8.22-gnu485
```

and I confirmed i can excute `h5cc -show` and returned:

`gcc -L/public/software/mathlib/hdf5/1.8.22-gcc485/lib /public/software/mathlib/hdf5/1.8.22-gcc485/lib/libhdf5_hl.a /public/software/mathlib/hdf5/1.8.22-gcc485/lib/libhdf5.a -lz -ldl -lm -Wl,-rpath -Wl,/public/software/mathlib/hdf5/1.8.22-gcc485/lib`

but when i entered the R environment, I met the error when installing BPCells :
```
** using staged installation
Testing hdf5 by compiling example program...

Retrying with pkg-config flags...
Error running 'pkg-config hdf5 --cflags --libs'

Searching for config information with h5cc -showconfig and h5cc -show...

Searching for hdf5 in a conda env...

Unable to locate libhdf5. Please install manually or edit compiler flags.
ERROR: configuration failed for package ‘BPCells’

```

In our HPC, we set hdf5 in a non-standard path. I have no idea what should I do to go on, although i found there's a similar issue and reply [https://github.com/bnprks/BPCells/issues/84](url) however it seems a bit different.

Thanks in advance.
Jacob Y. FU

yifanfu01:
Hi,
I solved this problem temporarily by editing the configure file in the source files:

First I load hdf5 environment, and use `h5cc -show` to find the pathway of hdf5 library, or `PATHWAY`.

Then I download the package and upload the source zip file to HPC. And change the configure files as follow:

(1) Change Lines 24-26 to 
CFLAGS="-I/PATHWAY/include"
CXXFLAGS="-I/PATHWAY/include"
LDFLAGS="-L/PATHWAY/lib"
HDF5_CFLAGS="-I/PATHWAY/include"
HDF5_OK="yes"

**AND**

(2) Delete Lines 34

Then I finally finished the installation.

Meanwhile, If guys forget deleting Line 34, the HDF5_CFLAGS will seems to be reset and will occur the error:
```
"In file included from arrayIO/../lib/highfive/H5DataSpace.hpp:18,
[...]

Issue #84, opened 2024-04-12
Install was not friendly when HDF5 was not installed via RPM in standard location


kgweiss:
Greetings,

One of my users requested help installing BPCells on our cluster.  We use LMOD modules to manage our software. The install configure for BPCells was having a difficult time finding my HDF5 installation. We used the .R/Makevars file and were able to install BPCells after providing the following compiler flags:

CFLAGS=-I<path to hdf5 install>/include
CXXFLAGS=-I<path to hdf5 install>/include
LDFLAGS=-L<path to hdf5 install>/lib
HDF5_CFLAGS=-I<path to hdf5 install>/include
HDF5_LIBS=<path to hdf5 install>/lib
LD_LIBRARY_PATH=<path to hdf5 install>/lib
INCLUDE=<path to hdf5 install>/include

FYI, we are using HDF5 version 1.10.8 and GCC version 10.3.0.

Just wanted to pass this along in case other users encounter install issues.

Cheers,
Ken

bnprks:
Hi Ken, thanks for the report! I'll also link #44 here, which might have some useful related discussion. One of the issues in that thread was that `LD_LIBRARY_PATH` on their cluster setup had more than one hdf5 version present which caused dynamic linking errors. The [`configure`](https://github.com/bnprks/BPCells/blob/main/configure) file has the logic for finding hdf5 versions.

In most cases I've seen, hdf5 issues are caused by some slightly non-standard setup on the user's end, such as a conda environment messing with library path resolution. Running `Sys.setenv(BPCELLS_DEBUG_INSTALL="true")` before installation will help print out what test-compilation commands are being run with any errors that are popping up. 

But in the event of issues, your approach of modifying `~/.R/Makevars` is a good way to brute-force R finding the correct library paths. Just FYI, I think only setting CFLAGS, CXXFLAGS, and LDFLAGS should be sufficient for BPCells but setting some additional ones as you did probably doesn't hurt.

PR #124, opened 2024-09-03
enable setting HDF5_CFLAGS and HDF5_LIBS external to ./configure

douglasgscofield:
The local `./configure` for BPCells does not have a mechanism for setting `HDF5_CFLAGS` and `HDF5_LIBS` external to the script, as do other `./configure` scripts; it relies only on its own internal mechanisms for determining the flags for interacting with hdf5. We do not provide hdf5 in the standard locations, but rather provide it from an Lmod module, so the include files and libraries for hdf5 are not in standard system locations..

This just uses a little bashiness to set the compilation and link flag variables to current environment variables if these are provided, otherwise it uses the current defaults.  With these changes, compilation against an hdf5 in a nonstandard location works correctly with

```
HDF5_CFLAGS="-I${HDF5_INCLUDE}" HDF5_LIBS="-L${HDF5_LIB} -lhdf5" R CMD INSTALL BPCells_0.2.0.tar.gz
```

assuming `HDF5_INCLUDE` is set to the hdf5 `.../include` directory and `HDF5_LIB` is set to its `.../lib` directory. Any other mechanism that can provide settings for these variables can be used to set these variables within the `./configure`.

This can address issues in #86 and #84.

### DIFF
--- a/r/NEWS.md
+++ b/r/NEWS.md
@@ -24,6 +24,7 @@ Contributions welcome :)
 - `trackplot_loop()` now accepts discrete color scales
 - `trackplot_combine()` now has smarter layout logic for margins, as well as detecting when plots are being combined that cover different genomic regions. (pull request #116)
 - `select_cells()` and `select_chromosomes()` now also allow using a logical mask for selection. (pull request #117)
+- BPCells installation can now also be configured by setting the `LDFLAGS` or `CFLAGS` as environment variables in addition to setting them in `~/.R/Makevars` (pull request #124)
 
 ## Bug-fixes
 - Fixed error message when a matrix is too large to be converted to dgCMatrix. (Thanks to @RookieA1 for reporting issue #95)
--- a/r/configure
+++ b/r/configure
@@ -1,8 +1,9 @@
 #!/bin/bash
+
 if [ -z $BPCELLS_DEBUG_INSTALL ]; then 
-    ERR=/dev/null
+    exec 3>/dev/null
 else
-    ERR=/dev/stdout
+    exec 3>&1
     set -x
 fi
 
@@ -21,9 +22,11 @@ fi
 CC=$("${R_HOME}/bin/R" CMD config CC)
 CXX=$("${R_HOME}/bin/R" CMD config CXX)
 
-CFLAGS=$("${R_HOME}/bin/R" CMD config CFLAGS)
-CXXFLAGS=$("${R_HOME}/bin/R" CMD config CXXFLAGS)
-LDFLAGS=$("${R_HOME}/bin/R" CMD config LDFLAGS)
+ENV_CFLAGS="${CFLAGS-}"
+ENV_LDFLAGS="${LDFLAGS-}"
+CFLAGS="$("${R_HOME}/bin/R" CMD config CFLAGS) $ENV_CFLAGS"
+CXXFLAGS="$("${R_HOME}/bin/R" CMD config CXXFLAGS) $ENV_CFLAGS"
+LDFLAGS="$("${R_HOME}/bin/R" CMD config LDFLAGS) $ENV_LDFLAGS"
 ############################
 # HDF5 compatibility check
 ############################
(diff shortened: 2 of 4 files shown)


bnprks:
Hi @douglasgscofield, thanks for taking a look at this - getting HDF5 to link properly can definitely be a challenge on some environments. Could you let me know a bit more about what the circumstances are of your cluster that require these changes as opposed to using the existing config options? 

(in order attempted, these are the R Makevars/`R CMD config` options; `pkg-config` options; `h5cc` autodetect; then `conda` autodetect. These try to start with the most standard approach, then fall back on less standard options if needed)

In your installation setup, will an R user always have the same HDF5 version (e.g. if the R lmod module depends on HDF5)? If so, simply adjusting the per-user or system-wide Makevars should be possible, as I described in [my comment](https://github.com/bnprks/BPCells/issues/86#issuecomment-2091230738) on #86. Due to some sloppiness on my part, I think it currently requires setting both `CFLAGS` and `CXXFLAGS` (and likely `LDFLAGS`).

Presumably your hdf5 lmod setup also does not set up `pkg-config` `.pc` files, or else the config would also be working already.

I can see how being able to set environment variables could make the config easier, though I'd have a slight preference for piggybacking on the more-standard `CFLAGS`, `CXXFLAGS`, `LDFLAGS` (or `LIBRARY_PATH`) variables rather than having the user set `HDF5_*` variables specific to BPCells only.

douglasgscofield:
Our build of hdf5/1.14.0 is from source, and that did not provide pkg-config support.

The pull request is meant to be in the spirit of `autotools`' `./configure` scripts, which have options like `--with-hdf5=$HDF5_PREFIX` or `cmake`'s `-D` variables for library and include file locations, both of which have cascading effects onto `CFLAGS`/`CXXFLAGS`/`LDFLAGS`.  If an option would be preferable to external environment variables that could certainly be done.

The hdf5 against which BPCells is built should be selectable at build time, since there is no base dependency of R on hdf5 and BPCells does not use `rhdf5` or another HDF5-facing R package that would introduce its own dependency. Meaning that the version built against is determined by our loaded module and the options to `./configure` as it is when we build against any other library.  Different users should be able to build against different hdf5 versions so long as their APIs are compatible with BPCell's expectations. RPATH info is sufficient (with `-L`options) to find the appropriate `libhdf5.so` at runtime.  Unfortunately things like `CPATH` and `LD_RUN_PATH` specified by our hdf5 module (and other library modules) seem rarely to be honoured during `./configure` and `cmake` runs, hence the necessity of library-specific options for locating these dependencies.

What is specifically happening with BPCells' `./configure` is that the system-installed hdf5 (1.10.x) is being found during the first test compilation, resulting in only `-lhdf5` being added to `LDFLAGS`, but we want to build against the newer module hdf5/1.14.0. This situation often occurs when clusters are running older OSs but we want to use newer libraries. My `./configure`and `cmake` lines can get quite long specifying dependencies.

bnprks:
Okay, that makes sense for why the environment variable would be preferred then. 
I take it that you are in the position of an admin who would have the ability to make the hdf5 modules modify some arbitrary environment variables in order to make BPCells installation run more smoothly across the cluster? 

Environment variable seems better for this than adding a command line option, as any users installing via `remotes::install_github` will more easily be able to set environment variables to alter the compilation.

My ideal case would be if the BPCells installation naturally respected `CPATH` and `LIBRARY_PATH` environment variables. From some quick testing it appears that `R CMD install` works with `CPATH` already, but `LIBRARY_PATH` didn't obviously do anything.

As a second choice, I'd probably prefer taking `CFLAGS` and `LDFLAGS` from the environment, so that we wouldn't have to add separate `HWY_CFLAGS` or `EIGEN_CFLAGS` for all other C/C++-level dependencies. Something along the lines of:
```
CFLAGS="$CFLAGS $("${R_HOME}/bin/R" CMD config CFLAGS)"
LDFLAGS="$LDFLAGS $("${R_HOME}/bin/R" CMD config LDFLAGS)"
```
but with some modification to make things propagate correctly to `Makevars.in` via that final `sed` command

I am on vacation for the next 10 days and won't be able to do much more checking here until then, but I will follow back up then

bnprks:
Hi @douglasgscofield, I'm back now from vacation. 

I'm currently leaning away from using your suggestion of reading `HDF5_CFLAGS` and `HDF5_LIBS` as environment variables, so I think you're off the hook for having to make code adjustments to satisfy my preferences. 

That said, I'd still be interested adjusting the `configure` script myself to address your use-case, perhaps with the solution I described above (reading `CFLAGS` and `LDFLAGS` from both the environment and R CMD config). I'd want to confirm with you that it would actually address your issue, though, before making the changes.

Let me know if that sounds like a reasonable plan

-Ben

douglasgscofield:
Does that mean that you would rewrite the `./configure` script to no longer use separate `HDF5` variables, it would all be run through `CFLAGS` and `LDFLAGS`?  That would allow a general mechanism for specifying options for dependencies and it seems like it would work for me.  So if you choose to go ahead with that, that seems reasonable.

Though I wonder if you are losing some specificity that would be helpful? Maybe reducing it to a single variable?  For example, CMake policy CMP0074 being true has it utilise a variable like `HDF5_ROOT` if it finds it, treating it like a prefix of an install tree, so if `./configure` finds `HDF5_ROOT` defined (and similarly for any other dependencies) and that location passes sanity checks, then `-I${HDF5_ROOT}/include` is added to `CFLAGS` and `-L${HDF5_ROOT}/lib -lhdf5` to `LDFLAGS`.  This is also analogous to a classic autotools option like `./configure --with-hdf5=$HDF5_ROOT ...`

Just an alternative that might be simpler, but I'm fine either way.

bnprks:
I'd probably keep most of the variables as-is in the configure script, mostly just add more flexibility to the user-facing configuration options. As a user, this would mean that in addition to setting `CFLAGS` and `LDFLAGS` in the standard `~/.R/Makevars` file, one could choose to use environment variables to provide flags to the compiler or linker respectively. BPCells would concatenate the flags read from both locations, and use those flags during all test compilation steps. (probably putting the environment variables second so they take priority)

My main concern here is keeping the user configuration options as simple to explain as possible, so I prefer allowing the user to directly set compiler flags rather than a more indirect method like specifying `HDF5_ROOT`.

If you don't mind me pushing some commits onto your personal main branch I can make my suggested edits in this PR, otherwise I can make a new demo branch to make sure it works for you before merging into main.

douglasgscofield:
Sure thing

bnprks:
Sorry for the delay on this, I've just pushed some updates to this PR that I think should address your issue. The configure script will now read + use any flags given in the environment variables `CFLAGS` and `LDFLAGS`.

While I was at it, I also made a change to fix the printing when running with the environment variable `BPCELLS_DEBUG_INSTALL` set. The only relevant bits for this fix though are the `ENV_CFLAGS` and `ENV_LDFLAGS` variables.

bnprks:
Hi @douglasgscofield, hopefully these changes are useful to you. I'm merging them in now to wrap this up, but feel free to file an issue or something if it doesn't provide the flexibility you need for your installation configuration

PR #128, opened 2024-09-18
[r][cpp] add pseudobulking to matrices

immanuelazn:


immanuelazn:
Added in the following changes

- R interrupt checking as well as removal of Rcpp code from physical logic has been finished for `pseudobulk_matrix()` and `matrix_quantile_per_cell()`.  These functions now follow the same styling as other functions, with their own cpp and header files.
- `pseudobulk_matrix()` calculates everything within a single pass, rather than separate passes for each aggregation method.  Using inspiration from `calculateMatrixStats()`
- `pseudobulk_matrix()` now returns a named list built from a struct, as we calculate the simpler aggregation methods during our pass anyways.
- `matrix_quantile_per_cell()` has been restricted to usage in non transposed matrices.  Should the user call this on a transposed matrix, they will be advised to call `transpose_storage_order()` first.
- Simple typo fixes and docstring clarity changes

r/tests/testthat/test-pseudobulk.R
```diff
@@ -0,0 +1,82 @@
...
+    # Check across two equal groups, one group, and groups of equal length
+    m <- matrices[[1]]
+    m_bpcells <- matrices[[2]]
+    groups <- c(rep.int(1, ncol(m)/2), rep.int(2, ncol(m)/2))
+    groups_one_type <- c(rep.int(1, ncol(m)))
+    groups_equal_length <- seq(ncol(m))
+    for (cell_group in list(groups, groups_one_type, groups_equal_length)) {
+      # create clipping
+      min_vals  <- m %>% apply(2, quantile, 0.99, type = 1)
+      #pmin doesn't work correctly when comparing matrix to vec, so we do it iteratively
```
bnprks, line 42:
```suggestion
      # create clipping
      min_vals  <- m %>% apply(2, quantile, 0.99, type = 1) %>% unlist()
      m_clipped <- t(pmin(t(m), min_vals))
      expect_false(m == m_clipped) # Check that clipping changes the values
```

r/tests/testthat/test-pseudobulk.R
```diff
@@ -0,0 +1,82 @@
...
+      m_bpcells_mean <- pseudobulk_matrix(m_bpcells, cell_group, method = "mean")$mean
+      m_bpcells_non_zeros <- pseudobulk_matrix(m_bpcells, cell_group, method = "non-zeros")$non_zeros
+      expect_equal(m_sum, m_bpcells_sum)
+      # check for clipping if matrix isn't transposed
+      if (m_bpcells@transpose == FALSE) {
+        m_clipped <- m
+        m_sum_clipped <- create_pseudobulk_r(m, cell_group, "sum")
+        m_bpcells_clipped <- pseudobulk_matrix(m_bpcells, cell_group, method = "sum", clip_values = TRUE)$sum
+        expect_equal(m_sum_clipped, m_bpcells_clipped)
+      }
```
bnprks, line 56: This is not testing the clipping, it's showing that the result with `clip_values=TRUE` is the same as without any clipping. All the reference values are computed off of `m` with no clipping applied.

bnprks (commented):
I've gotten through a detailed review pass, and ended up picking up on a lot of things I didn't notice in the first pass for high level feedback. 

I think the most important change to think about is how to make the R interface of `pseudobulk_matrix` a nice as we can, particularly with my comment about return values of list vs. raw matrix return. We should agree on what the right solution for it is, but I put some of my thoughts for options in the comment.

There were a couple other efficiency comments scattered through the C++, happy to chat more if the rationale for any of those is unclear or if how to implement is uncertain.

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,69 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+#'  - `mean`
+#'  - `var`
+#' @param clip_values (logical) If TRUE, clip high values to the 99th percentile of each cell.
+#' @param threads (integer) Number of threads to use.
+#' @return (Named list) A list of matrices with each matrix representing a pseudobulk matrix with a different aggregation method.
+#' Each matrix is of shape (features x groups), and names are one of `non_zeros`, `sum`, `mean`, `var`.
+#' @details The stats are ordered by complexity: nonzero, sum, mean, then variance. All
+#' less complex stats are calculated in the process of calculating a more complicated stat.
+#' So to calculate mean and variance simultaneously, just ask for variance,
+#' which will compute mean and nonzero counts as a side-effect.
```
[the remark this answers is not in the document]
immanuelazn: I think I would agree with number 1 more.  It would be relatively easy to implement and a little bit more intuitive

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,69 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+  }
+  return(res)
+}
+
+#' Find the nth quantile value of each cell in a matrix.
+#' @param quantile (numeric) quantile value to be found from each cell, between [0, 1].
+#' @return (Numeric) cell quantile values with number of entries equal to number of colums in the matrix.
+#' @inheritParams marker_features
+#' @export
+matrix_quantile_per_cell <- function(mat, quantile = 0.99) {
```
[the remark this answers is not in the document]
immanuelazn: I would agree to making it less single cell specific, and throwing it as a general matrix functionality.  We can do it like how `DelayedMatrixStats()` does it, however we should consider adding `rows`, `cols`, and `na.rm` subsetting

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,69 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+#' @param threads (integer) Number of threads to use.
+#' @return (Named list) A list of matrices with each matrix representing a pseudobulk matrix with a different aggregation method.
+#' Each matrix is of shape (features x groups), and names are one of `non_zeros`, `sum`, `mean`, `var`.
+#' @details The stats are ordered by complexity: nonzero, sum, mean, then variance. All
+#' less complex stats are calculated in the process of calculating a more complicated stat.
+#' So to calculate mean and variance simultaneously, just ask for variance,
+#' which will compute mean and nonzero counts as a side-effect.
+#' @inheritParams marker_features
+#' @export
+pseudobulk_matrix <- function(mat, cell_groups, method = c("sum", "mean", "var", "non-zeros"), clip_values = FALSE, threads = 1L) {
```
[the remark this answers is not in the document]
immanuelazn: Ah, I originally had the name `group` to match with `marker_features` but ended up changing it so the automatic documentation wasn't done.

I think your other points are completely fair.  The fourth point I think was a result of sum being used for mean calculations initially before changing this into a single pass solution.  It is an easy fix though.

immanuelazn:
change `matrix_quantile_per_cell()` to S3 generalizable `colQuantile()` function
change `colQuantile()` to use type 7 quantile calculation
change `pseudobulk_matrix()` to use a numeric clip_values representing quantile rather than boolean set to .99
change `pseudobulk_matrix()` to return a single matrix rather than a named list when only one method given
remove `matrix_quantile_per_row()`
fix problem with sum calculation in `pseudobulk_matrix()` when requesting more complex method
various documentation changes

r/src/bpcells-cpp/matrixUtils/Quantile.cpp
```diff
@@ -0,0 +1,98 @@
...
+                } else if (index < it.rows() - 1) {
+                    // there are no zeroes, and index is the last negative value
+                    col_num = curr[index] * (1-h) + curr[index + 1] * h;
+                } else {
+                    //there are no zeroes, and index is the last value
+                    col_num = curr[index];
+                }
+            } else {
+                // regular case for calculating quantiles
+                col_num = curr[index] * (1-h) + curr[index + 1] * h;
```
immanuelazn, line 60: kind of looking like a hellscape of if-else statements... any idea how I can clean this up a little?

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,102 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+#' feature.
+#' @param cell_groups (Character/factor vector)  of cell groups/clusters. Length must be equal to the number of cells.
+#' @param method (Character vector) Method(s) to aggregate counts. If one method is provided, the output will be a matrix. If multiple methods are provided, the output will be a named list of matrices.
+#' 
+#' Current options are:
+#'  - `nonzeros`
+#'  - `sum`
+#'  - `mean`
+#'  - `var`
+#' @param clip_values (integer) If set, clip matrix values to the `<clip_values>`th quantile of each cell, between 0 and 1 inclusive.
```
immanuelazn, line 85: Two notes:

- I left clip_values here, but I'm still open to changing it into a completely different function.  I left it here so I could test it together with `pseudobulk_matrix()` a little easier.  Any thoughts?
- If we do leave it here, we should probably change the name of the arg.  Any idea for something more intuitive?

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,102 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+
+#' Find the nth quantile value of each column in a matrix.
+#' @param x IterableMatrix object or a matrix-like object.
+#' @param probs (Numeric) Quantile value(s) to be computed, between 0 and 1.
+#' @return Numeric vector of quantile values for each column.
+#' @describeIn IterableMatrix-methods Calculate colQuantiles (replacement for `matrixStats::colQuantiles`)
+#' @export
+colQuantiles <- function(x, probs = 0.99, ...) UseMethod("colQuantiles")
+#' @export
+colQuantiles.default <- function(x, probs = 0.99, ..., na.rm = FALSE, type = 7) {
```
immanuelazn, line 143: I've noticed that all the functions that try to utilize the S3 dispatching `colMaxs()` `colVars()` etc all have non-related information in them.  We should discuss how to fix that!

r/R/transforms.R
```diff
@@ -254,7 +254,7 @@ min_by_col <- function(mat, vals) {
   }
   assert_is(mat, "IterableMatrix")
   assert_is(vals, "numeric")
-  assert_greater_than_zero(vals)
+  # assert_greater_than_zero(vals)
```
immanuelazn, line 257: Any zero values would cause this to fail.  Is this intentional?

bnprks (commented):
Here's a second round of changes. Good job overall with the various performance changes requested from last round. I commented a couple spots for further refinements within those changes you made.

One style change I've noticed a few times (perhaps copied from my own bad habits) is making a separate variable to track a `vector.size()`. This often just introduces more ways for things to go wrong. And I recently found from the docs that calling `vector.resize()` to a smaller size is guaranteed not to cause extra allocations/frees, so even when using a `vector` as just a memory buffer, it's always fine for performance to resize it down to the actual used size.

Other than that, I've been starting to get the hang of checking the generated assembly for what optimizations happened -- would be cool to demo that for you some time.

r/R/transforms.R
immanuelazn: I haven't resolved this, but are you saying that we just need some extra logic to skip a number if zero is iterated on, in `minbycol()`?

r/src/bpcells-cpp/matrixUtils/Pseudobulk.cpp
```diff
@@ -0,0 +1,112 @@
...
+        if (cell_group >= group_count_vec.size()) {
+            group_count_vec.resize(cell_group + 1);
+        }
+        group_count_vec[cell_group]++;
+    }
+    Eigen::Map<Eigen::Array<double, 1, Eigen::Dynamic>> group_count(group_count_vec.data(), group_count_vec.size());
+    // If transposed, also keep matrix transposed with groups as rows for cache-friendliness
+    uint32_t num_rows = transpose ? group_count_vec.size() : it.rows();
+    uint32_t num_cols = transpose ? it.cols() : group_count_vec.size();
+    if ((method & PseudobulkStatsMethod::NonZeros) == PseudobulkStatsMethod::NonZeros) res.non_zeros = Eigen::ArrayXXd::Zero(num_rows, num_cols);
```
immanuelazn, line 40: To allow for the bitwise checks, I just ended up comparing equality to the enum itself,  is this what you had in mind?

r/src/bpcells-cpp/matrixUtils/Pseudobulk.cpp
```diff
@@ -0,0 +1,112 @@
...
+        const uint32_t col = it.currentCol();
+        if (user_interrupt != NULL && *user_interrupt) return res;
+        while (it.load()) {
+            const uint32_t *row_data = it.rowData();
+            const T *val_data = it.valData();
+            const uint32_t count = it.capacity();
+            // Do transpose logic to calculate row_idx, col_idx out of loop, to avoid repeating conditional checks
+            if (transpose) {
+                // Doing variance calculation also does mean and non-zero count, so specific order is required to make sure we don't double count
+                if ((method & PseudobulkStatsMethod::Variance) == PseudobulkStatsMethod::Variance) {
```
immanuelazn, line 55: I reordered calculation so it is Variance -> Mean -> NonZeros. This allwoed me to remove some conditionals.

r/src/bpcells-cpp/matrixUtils/Quantile.cpp
```diff
@@ -0,0 +1,107 @@
...
+    std::vector<T> curr;
+    // clamp quantile, alpha, beta to [0,1]
+    for (auto& q: quantile) q = std::min(1.0, std::max(0.0, q));
+    alpha = std::min(1.0, std::max(0.0, alpha));
+    beta = std::min(1.0, std::max(0.0, beta));
+    // stats used for quantile index calculation
+    // store the index and gamma for each quantile
+    std::unordered_map<double, uint32_t> indexes;
+    std::unordered_map<double, double> gammas;
+    for (double& q: quantile) {
```
immanuelazn, line 60: This follows the scipy implementation, rather than the base R implementation

r/tests/testthat/test-pseudobulk.R
```diff
@@ -0,0 +1,147 @@
...
+  expect_error(pseudobulk_matrix(m1, cell_group, method = c("nonexistent_method", "sum")))
+})
+
+test_that("Matrix col quantiles works", {
+  m0 <- generate_sparse_matrix(20, 10)
+  m1 <- m0 |> as("IterableMatrix")
+  m0 <- as.matrix(m0)
+  # For single quantiles
+  for (quantile in c(0, 0.25, 0.5, 0.75, 0.99)) {
+    for (type in c(4, 5, 6, 7, 8, 9)) {
```
immanuelazn, line 117: I added additional tests to check for all types, to ensure the implementation for quantile is correct

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,201 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+    } else {
+      colnames(res[[res_slot]]) <- levels(cell_groups)
+      rownames(res[[res_slot]]) <- rownames(mat)
+    }
+  }
+  return(res)
+}
+
+
+#' Find the nth quantile value(s) of each row in a matrix. Only supports transposed matrices.
```
immanuelazn, line 136: I noticed that I can't add params in, or else the `IterableMatrix methods` page gets all screwed up.  It is probably a good time to start splitting that page up into a new section

immanuelazn:
Few notes:

- Added `rowQuantiles()`
- Change `colQuantiles()` to return a matrix if multiple probs are given
- Change `colQuantiles()` to work with any of the continuous quantile algorithms (4-9)
- Added tests for checking other types of quantiles, overall consolidation of quantile tests
- Changed `PseudobulkStatMethod` to use bitlags
- Changes `pseudobulk_matrix()` to not check the transpose conditional with every value check in the matrix.  Also cleaned up the code and incorporated aforementioned bit flags

r/tests/testthat/test-pseudobulk.R
```diff
@@ -0,0 +1,147 @@
...
+        }
+        if ("nonzeros" %in% methods_of_interest) {
+          expect_equal(m_non_zeros, m_bpcells_res$nonzeros)
+        }
+      }
+    }
+  }
+  # Check that the function errors out when an invalid method is passed
+  expect_error(pseudobulk_matrix(m1, cell_group, method = "nonexistent_method"))
+  expect_error(pseudobulk_matrix(m1, cell_group, method = c("nonexistent_method", "sum")))
```
bnprks, line 108: For these `expect_error` tests, it's often helpful to fill in the 2nd argument (regex on what the error message should say). In this case, you did have an error but it was uncontrolled rather than printing out something useful so the tests gave some false confidence.

r/tests/testthat/test-pseudobulk.R
bnprks: Looks good! Very cool we can support all of these types now

bnprks (commented):
Overall comments:
- I think we should ditch the clipping option. I think the ideal clipping for pseudobulks would measure per-gene quantiles within each pseudobulk, then apply the clipping within each pseudobulk.
  - A non-ideal but acceptable option would be to clip along rows. This at least applies the clipping per-gene, but might cause issues when a rare cell population expresses a certain gene much higher than other cells. Then the clipping might eliminate the high expression of that marker.
- You should remember to worry about how invalid arguments get handled. You don't need to add unit tests for these, but I like to manually run a small example to check nothing bad happens if I pass a bad value to each argument in sequence. (See the [code polishing checklist](https://github.com/bnprks/BPCells/wiki/Code-polishing-checklist))
  - In general, I find doing a little bit of manual testing is a good way to check error messages and edge cases more quickly. Those don't always require added tests (particularly edge cases that turn out to work correctly without any special handling in the code)
  - Prefer col-major or row-major over saying "non-transposed" or "transposed". This is a little bit of a mismatch between the internal variable names BPCells uses and the external communication. Functions like `storage_order` and `show` return the row or col major language, and error messages should match that too.

I've also put these follow-up tasks into the initial description of this PR so we can remember to come back:
- [ ] Move row/colQuantiles into a different file. Perhaps make a `matrixStats.R` file that collects all of the matrixStats interface functions
- [ ] Split function documentation to have the `matrixStats` interfaces as a page, and then maybe split up the arithmetic functions into a few groups too

r/R/singlecell_utils.R
```diff
@@ -68,4 +68,201 @@ marker_features <- function(mat, groups, method="wilcoxon") {
...
+#' @export
+rowQuantiles.IterableMatrix <- function(x, rows = NULL, cols = NULL,
+                                        probs = seq(from = 0, to = 1, by = 0.25),
+                                        na.rm = FALSE, type = 7L, digits = 7L, ...,
+                                        useNames = TRUE, drop = TRUE) {
+  if (!is.null(rows) || !is.null(cols) || isTRUE(na.rm) || (digits != 7L) || isFALSE(useNames) || isFALSE(drop)) {
+    rlang::abort("rowQuantiles(IterableMatrix) doesn't support extra arguments rows, cols, na.rm, digits, useNames, or drop.")
+  }
+  if (!x@transpose) {
+    rlang::abort("rowQuantiles(IterableMatrix) only supports transposed matrices. Please call transposed_storage_order() first, then run colQuantiles.")
```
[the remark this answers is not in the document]
immanuelazn: I think I'm having a fundamental misunderstanding on the second point, wouldn't calling transpose_storage_order() cause the matrix to be stored row-major, ie in transposed state? In this case, calling `colQuantiles()` is what the user would want, right?

immanuelazn:
I think I have addressed all comments, I also gave it a pass with manually checking everything.  As for the two points that you have put up, I put them into an issuue within the projects page, for a new PR.  Thanks for being so patient and detailed with your review Ben

r/src/bpcells-cpp/matrixUtils/Pseudobulk.cpp
```diff
@@ -21,26 +21,29 @@ namespace BPCells {
 template <typename T>
 PseudobulkStats pseudobulk_matrix(std::unique_ptr<MatrixLoader<T>> &&mat,
                                   const std::vector<uint32_t>& cell_groups,
-                                  PseudobulkStatsMethod method,
+                                  int method,
```
immanuelazn, line 24: Not entirely sure if it is 100% worth the syntactic cleanliness, if we have to pass these in now as ints.  I feel like this kind of obfuscates how to use the method argument.

bnprks (commented):
Sorry for the slow review, almost all there. 

The pseudobulk_matrix function looks completely good now. You're right that switching the C++ bitflag code was not obviously beneficial -- sorry for the in retrospect unhelpful suggestion.

There are a few small issues still on the rowQuantiles and colQuantiles interfaces which I've commented in place -- rowQuantiles isn't quite forwarding all args and colQuantiles is returning transposed relative to `matrixStats`.

immanuelazn:
- Changed quantile tests to use `matrixStats` functions, so we can ensure correct syntax and behaviour
- Reverted `PseudobulkStatsMethod` bit flagging to no longer be cast as ints.
- Transposed `colQuantiles()` to match `matrixStats`

bnprks (commented):
- [x] For quantile tests, need a `skip_if_not_installed("matrixStats")` call since `matrixStats` is only a test-time dependency so might not necessarily be installed
- [x] With the changed orientation of `matrix_quantile_per_col()`, the `Quantile.h` file comment needs updating to be accurate for the return value description
- [x] Line 50 of `Quantile.cpp` contains commented out code
- [x] `singlecell_utils.R` line 132 docstring description should say rowQuantiles for both references
- [x] `singlecell_utils.R` line 130, 193 need to prefix with `- FunctionName():` to render properly in the website docs
- [ ] Discussion: what file to put the quantile code in, and also what files to put tests in
   - Quantile tests -> `test_matrix_stats.R`
   - Quantile R implemenation -> initialize `matrix_stats.R`
   - Pseudobulk tests -> `test_singlecell_utils.R`

PR #137, opened 2024-10-08
[r] sklearn/LSI interface design docs

immanuelazn:


designs/sklearn-interface.md
immanuelazn, line 1: Any idea on the naming for the classes?  These definitely have to change, especially the SVD since we already have an implementation in BPCells

immanuelazn:
Closed as we are changing direction towards #167

Issue #140, opened 2024-10-17
error install BPCells, no matching function for call to 'min'


hpsunlearning:
I am trying to install BPCells on a Mac M1, and I figured out how to fix the hdf5 issue but another error popped out. 
has anyone encountered the same issue?

bitpacking_io.cpp:259:51: error: no matching function for call to 'min'
  259 |         uint64_t loaded = reader.load(output + i, std::min(chunk_size, capacity-i));

<details>
  <summary><i>error output</i></summary>
> remotes::install_github("bnprks/BPCells/r")

Downloading GitHub repo bnprks/BPCells@HEAD
── R CMD build───────────────────────────────────────────────────
✔  checking for file ‘/private/var/folders/cs/lcn8f90s0jg124ff3y1twlkr0000gn/T/Rtmpr2SRhI/remotes117b9e6a9c9/bnprks-BPCells-fd72cb1/r/DESCRIPTION’ ...
─  preparing ‘BPCells’:
✔  checking DESCRIPTION meta-information ...
─  cleaning src
─  running ‘cleanup’
─  checking for LF line-endings in source and make files and shell scripts
─  checking for empty or unneeded directories
─  building ‘BPCells_0.2.0.tar.gz’

* installing *source* package ‘BPCells’ ...
** using staged installation
Testing hdf5 by compiling example program...

Retrying with pkg-config flags...
Found working hdf5
HDF5_CFLAGS='-I/opt/homebrew/Cellar/hdf5/1.14.4.3/include'
HDF5_LIBS='-L/opt/homebrew/Cellar/hdf5/1.14.4.3/lib -lhdf5'

Testing C++17 filesystem feature support...
Testing availability of highway SIMD library...
Building highway SIMD library from source
** libs
using C++ compiler: ‘Apple clang version 16.0.0 (clang-1600.0.26.3)’
using SDK: ‘MacOSX15.0.sdk’
[...]

PR #141, opened 2024-10-18
Fix broken Mac build and add CI checks

bnprks:
This is designed to fix issue #140 and prevent similar issues in the future by having checks run on new PRs.

It turns out there was more than one thing wrong with the builds, so the list of fixes is:
- Issue #140 caused by a `std::min()` call in `bitpacking_io.cpp`
- `tile_matrix(mode="fragments")` had a broken SIMD implementation that was only caught with 4-wide SIMD
- `sctransform_pearson()` had lower precision than it should have on ARM due to a typo when calculating Newton-Raphson iterations of `ReciprocalSqrt` for iterations >1
- Don't use an unoptimized `-O0` build for windows, as there is a strange compiler bug that gets triggered affecting SIMD code and causing segfaults. 
  - This appears to be a similar problem as [this highway issue](https://github.com/google/highway/issues/332)
  - In short, the crash is from performing an aligned load of an AVX vector argument from an unaligned stack address. (I confirmed this is happening in gdb). AVX requires 32-byte alignment, but the stack only guarantees 16 I believe. Optimized builds cause all these functions to be inlined so there is no issue reading vector arguments from the stack.

There remain two unfixed issues to follow up on. For now, the relevant tests have been marked to skip:
- Code related to peak calling uses `parallel::mclapply()` for parallelism which isn't supported on windows. 
  - Moving this parallelism into C++ would fix the problem. 
  - The calls to `macs` subprocesses might be tricky, particularly while having Ctrl-C get handled properly. the `processx` R library may have some useful learnings to supply (e.g. [this blog post](https://blog.r-hub.io/2020/02/20/processx-blocked-sigchld/))
- Maximum open file issues happen for me on the windows CI
  - This issue doesn't reproduce for me testing locally on my windows machine
  - This should be handled by `_setmaxstdio` calls in `binaryfile.h`, but maybe we can do some print debugging to see if those calls are failing in the CI environment.

A future follow-up goal to improve CI would be fixing all the skipped tests and adding test coverage tracking via `covr`


.github/workflows/r-test.yml
```diff
@@ -0,0 +1,49 @@
+name: Build and test R package
+
+on: [pull_request]
```
immanuelazn, line 3: I think it would be a good idea to consider whether we want this for all PRs, or for just PRs to main

r/src/bpcells-cpp/simd/current_target.cpp
```diff
@@ -34,5 +37,25 @@ const char* current_target() {
     return HWY_DYNAMIC_DISPATCH(current_target)();
 }
 
+
```
immanuelazn, line 40: Even though these are tiny functions, a little bit of a docstring never hurts!

r/tests/testthat/test-atac_utils.R
```diff
@@ -154,6 +154,7 @@ test_that("write_insertion_bedgraph works", {
 })
 
 test_that("write_insertion_bed works", {
+  skip_on_os("windows") # TODO: threads >1 causes parallel::mclapply issues on windows
```
immanuelazn, line 157: As discussed, we should set Windows to use 1 thread as a fallback for now, and get a fix through moving the multithreading to C++

r/tests/testthat/test-matrix_stats.R
```diff
@@ -190,6 +190,7 @@ test_that("svds registers generic with RSpectra", {
 })
 
 test_that("rowMaxs and colMaxs works comprehensive", {
+  skip_if_not_installed("matrixStats")
```
immanuelazn, line 193: As we will be growing the amount of functions that utilize matrixStats interfaces, I would suggest doing an install of `matrixStats` within the workflow itself.  It is already in the `suggests` within the R description

bnprks: The `r-lib/actions/setup-r-dependencies@v2` does already install all the `suggests` dependencies, so these will run. This change was just a drive-by change tangentially related to this PR.

immanuelazn:
Changes look good!  I'll prioritize doing the windows multithreaded insertion bedfile writing next.  Thanks Ben!

PR #143, opened 2024-10-22
[r] Change default insertion mode for macs peak calling, rename function

immanuelazn:
Deprecating `call_macs_peaks()` to `call_peaks_macs()` as discussed previously.  Note, is this how we should be handling deprecated APIs?


r/R/atac_utils.R
```diff
@@ -606,11 +606,18 @@ call_macs_peaks <- function(fragments, path,
   }
 }
 
+#' @export
+call_macs_peaks <- function(...) {
+  rlang::warn("`call_macs_peaks()` is deprecated. Please use `call_peaks_macs()` instead.")
+  return(call_peaks_macs(...))
+}
```
immanuelazn, line 613: Is this an adequate way of deprecating an interface?

bnprks (commented):
Overall looks good, just a couple small comments below. I'm assuming that we'll save working on the Windows C++ threading for another PR?

immanuelazn:
Cool, I'll make the changes with the better way of styling!

I think it would make sense to do the multithreading as a separate PR as that requires work more on the internal logic, rather than interface/documentation

immanuelazn:
Styling changes from deprecation finished!

bnprks:
Overall, this looks quite good and is ready to merge. 

Something we should discuss separately: do we knew why we missed the default argument issue initially despite comparing outputs against ArchR. Do you think our checks didn't cover this or weren't running with insertion_mode set to the default?

immanuelazn:
Yup, so the results were compared in this [file ](https://github.com/bnprks/BPCells/blob/334f172ea8e4b5efe81dd35a5b68a4dae03c13b7/r/tests/real_data/ArchR_insertions.R#L134).  I passed in the argument for `both` in `insertion_mode` within this test case but didn't change the default param on the macs, and bedfile themselves.

PR #144, opened 2024-10-23
[r] add automatic website building

immanuelazn:
Creates a PR to html-docs whenever something is pushed to `main`, and performs the website building workflow indicated in `Wiki/Tool Development Workflows`
Title of PR is `Update docs for commit: <most_recent_commit_message>`

Message of PR is the git log of the most recent commit.  I am open to changing this to something that better describes the changes to documentation.
Example of how this would look can be found [here](https://github.com/immanuelazn/BPCells/pull/17).  This PR was created as a result of pushing this commit into my personal BPCells main branch.

We can also make it perform an automatic push to `html-docs`, if that is preferred.  We can also set this up to be a manually triggered workflow.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -39,8 +39,8 @@ jobs:
         Rscript .github/workflows/deploy-reference.R
     - name: Configure Git
       run: |
-        git config --global user.name "GitHub Action"
-        git config --global user.email "action@github.com"
+        git config user.name 'github-actions[bot]'
+        git config user.email 'github-actions[bot]@users.noreply.github.com'
     - name: Create new branch
       run: |
         git checkout -b update-docs-$(git rev-parse --short=6 ${{ github.sha }})
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,62 @@
+name: Build Docs and Create a PR with Docs Changes
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build_docs:
+    name: Build and Deploy Documentation
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # to ensure other branches are available for build
+    - name: Install system dependencies
+      run: sudo apt-get install -y libhdf5-dev git-restore-mtime
+    - name: Set r compilation options
+      run: bash -c 'echo -e "MAKEFLAGS=--jobs=3\nCXXFLAGS += -O1 -UNDEBUG" > "$GITHUB_WORKSPACE/Makevars.user" && echo "R_MAKEVARS_USER=$GITHUB_WORKSPACE/Makevars.user" >> "$GITHUB_ENV"'
+    - name: Setup R
+      uses: r-lib/actions/setup-r@v2
+    - name: Install R dependencies
+      uses: r-lib/actions/setup-r-dependencies@v2
+      with: 
+        cache-version: 1
+        working-directory: 'r'
+        extra-packages: |
+            any::pkgdown
+            any::devtools
+            any::uwot
+            any::irlba
+            any::RcppHNSW
+            any::igraph
+            any::BiocManager
+            bioc::BSgenome.Hsapiens.UCSC.hg38
+            github::GreenleafLab/motifmatchr
+            github::GreenleafLab/chromVARmotifs
+    - name: Install BPCells
+      run: |
+        Rscript -e 'install.packages("r", repos=NULL, type="source")'
+    - name: Create git worktree with docs
+      run: |
+        git worktree add r/docs docs-html
+    - name: Build documentation
+      run: |
+        Rscript .github/workflows/deploy-reference.R FALSE
+    - name: Configure Git
+      run: |
+        git config --global user.name "GitHub Action"
+        git config --global user.email "action@github.com"
+    - name: Create new branch
+      run: |
+        git checkout -b update-docs-$(git rev-parse --short=6 ${{ github.sha }})
+        git add .
+        git commit -m "Update documentation for commit $(git rev-parse --short=6 ${{ github.sha }})"
+      working-directory: r/docs
+    - name: Push changes to new branch
+      run: git push origin update-docs-$(git rev-parse --short=6 ${{ github.sha }})
+    - name: Create Pull Request # use message with last commit
+      run: |
+        gh pr create --title "Update docs for commit: $(git log -n 1 --pretty=%B)" --body "$(git log -n 1 main)" --base docs-html --head update-docs-$(git rev-parse --short=6 ${{ github.sha }})
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -13,7 +13,7 @@ jobs:
       with:
         fetch-depth: 0 # to ensure other branches are available for build
     - name: Install system dependencies
-      run: sudo apt-get install -y libhdf5-dev git-restore-mtime
+      run: sudo apt-get install -y libhdf5-dev
     - name: Set r compilation options
       run: bash -c 'echo -e "MAKEFLAGS=--jobs=3\nCXXFLAGS += -O1 -UNDEBUG" > "$GITHUB_WORKSPACE/Makevars.user" && echo "R_MAKEVARS_USER=$GITHUB_WORKSPACE/Makevars.user" >> "$GITHUB_ENV"'
     - name: Setup R
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -23,10 +23,6 @@ jobs:
       with: 
         cache-version: 1
         working-directory: 'r'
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,58 @@
...
+      uses: r-lib/actions/setup-r-dependencies@v2
+      with: 
+        cache-version: 1
+        working-directory: 'r'
+        extra-packages: |
+            any::pkgdown
+            any::devtools
+    - name: Get last commit message
+      id: last_commit
+      run: |
```
immanuelazn, line 31: Super ugly way to do multi-line environment variables.  GH Actions docs themselves indicate to do it this way, but I'm open to doing something more clean if you have an idea

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,58 @@
...
+          echo 'EOF'
+        } >> "$GITHUB_ENV"
+    - name: Create git worktree with docs branch
+      run: git worktree add r/docs docs-html
+    - name: Build documentation
+      run: |
+        Rscript .github/workflows/deploy-reference.R
+    - name: Configure Git
+      run: |
+        git config --global user.name "${{ github.actor }}"
```
immanuelazn, line 44: Choice to create the documentation commit using the person who created the PR.  We can change this behaviour to the `gh actions` worker instead though

bnprks (commented):
This looks like it will be very helpful! I like the interface of having a PR created so there's a chance for manual review before deploying to the website.

I have a few more thoughts on options for handling the vignette rebuilds:
1. Just try re-building the vignettes and see how long it is and whether we get a lot of plot images repeatedly re-generated. Since this will kick off automatically I don't mind if it takes 30 minutes to run.
2. Try something fancy to avoid re-generating unchaged vignettes. E.g. saving the existing vignette outputs for unchaged vignettes and deleting the Rmd source, then restoring those files after re-generating the site with those files held out. Using a command like `system("git diff --name-only HEAD HEAD~1", intern=TRUE)` should be helpful here
3. Make a secondary workflow that runs a full website rebuild but is scoped to only run when a vignette file is updated ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files))

I think the re-build everything option is probably easier to implement and might work fine. Being fancier about not re-generating old vignettes is probably doable but kinda error prone. 

Another robustness concern I thought of is that we probably want our website rebuild to be able to delete files from the website too, not just add or modify. So that's something handled well by the `pkgdown::build_site_github_pages()` function.

immanuelazn:
Getting selective changes for vignettes was much more involved than anticipated.  Apparently, we can't use `pkgdown::build_articles()` with the approach of renaming the non-changed files' extensions, as this breaks the linking to index.  Additionally, using `lazy=TRUE` doesn't work with resetting git modification times off the box, as we use worktrees to build our website.  

I decided to build the selective article rendering with using `pkgdown::build_article()`, and that seems to work!
I also created a script that can manually be run, and redo's website building.

immanuelazn:
Demo of the new build process:

- I prd the workflow into my fork's main branch.  This auto triggered this [PR](https://github.com/immanuelazn/BPCells/pull/42).  I then manually ran the workflow for full website build in this [PR](https://github.com/immanuelazn/BPCells/pull/43)
- I did a trial commit with only small syntactic changes in two vignettes.  This triggered this [PR](https://github.com/immanuelazn/BPCells/pull/44/files), and reflects that changes were only made to two vignettes

.github/workflows/deploy-full-website.yml
```diff
@@ -0,0 +1,60 @@
...
+            any::pkgdown
+            any::devtools
+            any::uwot
+            any::irlba
+            any::RcppHNSW
+            any::igraph
+            any::BiocManager
+            bioc::BSgenome.Hsapiens.UCSC.hg38
+            github::GreenleafLab/motifmatchr
+            github::GreenleafLab/chromVARmotifs
```
bnprks, line 34: It looks like we can actually get these packages out of the workflow and into our `DESCRIPTION` file under a section `Config/Needs/website:`. See [r-pkgs book](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-nonstandard-config-needs) and [ggplot2 github action](https://github.com/tidyverse/ggplot2/blob/f12b73de2609603ff405360c73f1a5f38466af45/.github/workflows/pkgdown.yaml#L35-L38). From [searching github](https://github.com/search?q=%22Config%2FNeeds%2Fwebsite%3A%22+path%3ADESCRIPTION+&type=code), it appears that it's possible to list these one-per-line without commas, e.g. as [tidymodels/workflows](https://github.com/tidymodels/workflows/blob/main/DESCRIPTION) does.

bnprks (commented):
Overall this looks really good! I've added a couple inline comments.

Some other high-level comments:
- It might be possible to only have a single workflow that is triggered by push or workflow_dispatch, then making the actual docs build step conditional on  `if: github.event_name == "workflow_dispatch"` ([example from docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#example-usage-of-the-github-context)) ([example from ggplot2](https://github.com/tidyverse/ggplot2/blob/f12b73de2609603ff405360c73f1a5f38466af45/.github/workflows/pkgdown.yaml#L44-L50))
- Even with the fancy incremental build, I think we'll always be hard pressed to recreate the logic of `pkgbuild::build_site(lazy=TRUE)`. I really think it might make sense to do:
  ```
  git restore-mtime
  git restore-mtime --cwd=$(pwd)/r/docs
  Rscript -e 'pkgdown::build_site(lazy=TRUE)'
  ```
  Though we can discuss the tradeoffs a bit -- there's some pros and cons either way

.github/workflows/deploy-full-website.yml
immanuelazn: Apparently there's some added complexity for importing from non-CRAN sources.  I added everything else in that didn't have a `bioc` or `github` label into `DESCRIPTION` as suggested!

.github/workflows/deploy-reference.R
```diff
@@ -0,0 +1,21 @@
...
+args = commandArgs(TRUE)
+# First argument is whether to rebuild the entire site.
+rebuild <- as.logical(args[1])
+if (rebuild) {
+  pkgdown::build_site_github_pages('r')
+} else {
+  pkgdown::build_reference('r')
+  build_selective_vignette_changes()
+  pkgdown::build_news('r')
+  pkgdown::build_home('r')
```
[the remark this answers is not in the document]
immanuelazn: I ended up calling `build_sitemap()` even though it's an internal function.  I also reordered to follow the ordering of `build_site()` a little bit better, and also added the rest of the functions you suggested

immanuelazn:
Some comments:

- I ended up just sticking to the current infrastructure route, instead of taking the github mtime changing option.  
- I tested with the following workflow:
  - PRd the changes to my main branch (PR looks like[ this ](https://github.com/immanuelazn/BPCells/pull/77), but [this](https://github.com/immanuelazn/BPCells/pull/73) is what it looks like when I don't squash commits).  Both look acceptable.  Action reflecting the build found [here](https://github.com/immanuelazn/BPCells/actions/runs/11565041640)
  - I commited a change where I added the word "test" to the benchmarking Rmd, and added a new Rmd (Looks like [this](https://github.com/immanuelazn/BPCells/pull/78)).   Action for this change found [here](https://github.com/immanuelazn/BPCells/actions/runs/11565041640.) We can see the git diff reflects the articles being updated accordingly, when compared to the previous steps PR.  I also added this to personal website, and the articles are updated correctly (Found [here](https://immanuelazn.github.io/BPCells/articles/index.html)).
  - I also manually rebuilt the docs to confirm that its still works.  The non-deterministic nature is still an issue.  Action reflecting the change found [here](https://github.com/immanuelazn/BPCells/actions/runs/11565056887).  PR found [here](https://github.com/immanuelazn/BPCells/pull/79)
  - I then tested when I added non-linked png files, and can confirm that these do not create a PR.  I had to add a change to handle breaking early.  GH actions still perceives it as a failure (Action found [here](https://github.com/immanuelazn/BPCells/actions/runs/11565100598)).  People are not happy that breaking early is not conventionally supported in GHA (as shown [here](https://github.com/actions/runner/issues/662)).  There are ways around it, but  I don't think it's worth the additional complexitiy in the workflow.  In the end the workflow still does its job

bnprks:
This looks great! Should be very useful to help keep the docs site updated. Ready to merge from my perspective

For my own curiosity (we can discuss offline at some point)
- I noticed that you left some `bioc::` and `github::` packages off the `DESCRIPTION` `Config/Needs/website` section. I kind of expected those prefixes to just work since the github action ends up calling `pak` to do the install ([supported prefixes](https://pak.r-lib.org/reference/pak_package_sources.html)). You mentioned some added complexity if trying to use those -- what was it that came up?

immanuelazn:
Ah so that is my bad, I read that you need to have all external packages in a seperate header in your DESCRIPTION file.  I looked through a few public repos and I couldn't find any that actually imported bioc packages through the same website header. This thought process was incorrect, and manual testing shows that it does indeed work.  Good catch!  PR with full rebuild shown [here](https://github.com/immanuelazn/BPCells/pull/80)

Issue #36, opened 2023-07-05
load data


yuq1993:
Hi, 

Thank you for developing this tool. 
I am trying to use open_matrix_anndata_hdf5 to load anndata from a paper. I got below error:

Error in eval(expr, envir, enclos): Error in opening AnnData matrix: "X" is a dataset rather than a group. This likely indicates a dense matrix which is not yet supported by BPCells.

Does this mean this anndata is not compatible with BPCells? Is there a way to solve it?
My second question is can BPCells load in MEX-format gene-barcode counts from 10x output?

Thank you!
Best, 
Qian

bnprks:
Hi Qian, thanks for your questions. AnnData with sparse matrices (the most common case) is compatible with BPCells, but as described by this error message and issue #17, dense expression matrix formats are not yet supported.

You can likely convert the AnnData to use a sparse matrix format within python, using the SciPy sparse functionality ([tutorial](https://docs.scipy.org/doc/scipy/tutorial/sparse.html)) and the AnnData API.

For your second question, MEX-format import is supported in BPCells, via the `import_matrix_market` and `import_matrix_market_10x` functions ([link](https://bnprks.github.io/BPCells/reference/import_matrix_market.html)). For technical reasons this is an import (i.e. file format conversion) rather than a direct load, but it should work similarly from your perspective as an end-user.

Issue #17, opened 2023-04-07
Dense h5ad matrix to BPCells


yuhanH:
Hi,
I find that `open_matrix_anndata_hdf5` cannot open h5ad files from Genome-wide Perturb-seq datasets:
https://plus.figshare.com/articles/dataset/_Mapping_information-rich_genotype-phenotype_landscapes_with_genome-scale_Perturb-seq_Replogle_et_al_2022_processed_Perturb-seq_datasets/20029387

bnprks:
I checked out one small file quickly -- K562_essential_raw_bulk_01.h5ad, and it looks like it's storing the matrix in dense format rather than sparse. I'm not sure if this is true for the large files as well. 

Thanks for alerting me to this problem, but I don't think BPCells is going to add support for reading AnnData dense matrices in the near future (unless many more people chime in with examples of common datasets distributed in this manner)

My recommendation for a workaround would be either:
1. Read the AnnData file in python, convert to a scipy sparse matrix in memory, then write back out to AnnData with a sparse matrix
2. Read the "X" matrix in the file directly using an R package like hdf5r (my favorite) or rhdf5

yuhanH:
Yeah. You are right. The count matrix is saved as a dense matrix there. By converting it to a sparse matrix using scipy sparse, BPCells is able to read the h5ad file. Thank you!

yuhanH:
One following question. If the count matrix is stored at `adata.raw.X`, can `open_matrix_anndata_hdf5` read it directly?

bnprks:
Yeah, as listed in the [docs](https://bnprks.github.io/BPCells/reference/open_matrix_anndata_hdf5.html) just pass the correct `group` path separated by slashes. I'm assuming for your question that would be `raw/X` though you may need to dig a bit with `h5ls` to find the correct path in your particular file. (This is all assuming it's a sparse matrix format of course)

yuhanH:
Thank you for your help! `h5ls ` is working. When i set `group = '/raw/X' `, it can read that matrix by `open_matrix_anndata_hdf5 `

PR #146, opened 2024-10-23
Add AnnData dense matrix read support

bnprks:
This addresses old issues #17 and #36 by adding support for reading the AnnData dense matrix format.

Changes:
- Add support for reading dense AnnData matrices
- Adjust test files for old AnnData versions to also cover `obsm`, `varm`, and dense matrices. Reduce the version numbers covered since v0.7.8 appeared to be identical to v0.7.0
- Adjust how dimnames are recognized from AnnData files (match based on length rather than always assuming var is rows and obs is cols. This helps support other matrix shapes as found in `obsm`, or `varp`, etc.)
- Split the 10x and AnnData import code into separate files since they are getting a bit long otherwise

The first commit is the main one to look at for review, the others are moving around code mostly.


r/src/bpcells-cpp/matrixIterators/ImportMatrixAnnDataHDF5.cpp
```diff
@@ -557,33 +557,19 @@ std::string get10xMatrixType(std::string file, std::string group) {
...
-    } else if (g.hasAttribute("encoding-type")) {
-        g.getAttribute("encoding-type").read(encoding);
-    } else {
-        throw std::runtime_error(
-            "h5ad could not be read - missing attribute 'encoding-type' on sparse matrix"
-        );
-    }
+    std::vector<uint32_t> dims;
+    bool row_major, is_sparse;
+    readAnnDataDims(h5file, group, dims, row_major, is_sparse);
```
immanuelazn, line 570: I like how this is now a much cleaner function.  However, I think that using a void typed function to get the row_major status makes it a little bit unclear on how you obtained row_major status.  In terms of readability, having to look into the function call on `readAnnDataDims` to see that your values changed as they're passed by reference is an extra step.  Maybe we can just make it return the bool, or the tuple of bools?

bnprks:
As discussed offline, another to-do item is filtering out explicit zeros while reading from a dense matrix

r/tests/testthat/test-matrix_io.R
```diff
@@ -260,14 +264,24 @@ test_that("AnnData read backwards compatibility", {
     open_matrix_anndata_hdf5(file.path(dir, f), group="layers/transpose") %>%
       as("dgCMatrix") %>%
       expect_identical(ans)
+    open_matrix_anndata_hdf5(file.path(dir, f), group="layers/dense") %>%
+      as("dgCMatrix") %>%
```
immanuelazn, line 268: I think it would be good to check for writing matrices that were originally read dense.  I did a little bit of testing with `write_matrix_hdf5()`, and `write_matrix_memory()` and they appear to be working.  But it never hurts to have a check in case of a regression!

immanuelazn (approved):
Overall looks super good! Admittedly I had to read through a lot of highfive APIs, as well as through AnnData, to get a full understanding, but this looks super clear.  Runs on my end well!

r/tests/testthat/test-matrix_io.R
bnprks: Since `as("dgCMatrix")` calls `write_matrix_memory()` internally, I think this is already covered here

r/src/matrix_io.cpp
```diff
@@ -70,6 +70,11 @@ template <class T> List dims_matrix(StoredMatrix<T> &&mat, bool transpose) {
     }
 }
 
+template <class T> List dims_matrix(std::unique_ptr<MatrixLoader<T>> &&mat, bool transpose) {
+    // This is only safe because we know the main `dims_matrix` function doesn't store any references to the object
+    return dims_matrix(std::move(*mat), transpose);
+}
```
bnprks, line 76: I need to do a bit more research on the safety of this before merging. I was just running some small C++ tests and getting surprised by when destructors got called, so my mental model is not quite right

bnprks: I think I have figured this out now. The solution I used was:
- Make the base `dims_matrix` take an l-value reference, since that provides more ownership guarantees to the caller (i.e. the object will not get consumed by the function call)
- Make a helper overload that takes an r-value reference. This provides fewer guarantees to the caller but is otherwise safe
- Dereferencing a std::unique_ptr with `*` produces an l-value, which now can be passed directly to the l-value reference overload of `dims_matrix`

immanuelazn:
No qualms about the new changes you added!  Looks good Ben.

Issue #125, opened 2024-09-04
BPCells::svds Failing on Large Matrix


Sandman-1:
Hello! Thank you so much for this package. I would like to run svds using the BPCells wrapper function on a matrix containing 9,096 genes and 4,188,533 cells. I followed the workflow outlined in the vignette, including writing a temporary matrix file using write_matrix_dir(scaled, tempfile("mat")) before running BPCells::svds on this temporary matrix. Unfortunately, it returned the following error after ~40 hours of runtime: Error: TridiagEigen: eigen decomposition failed. Could you please help me resolve this issue?

`

- srt <- FindVariableFeatures(srt, nfeatures = round(0.25*nrow(srt))) 
- cell_peak_stats <- matrix_stats(srt@assays$RNA$data, col_stats="variance")$col_stats 
- cell_means <- cell_peak_stats["mean",] 
- cell_vars <- cell_peak_stats["variance",] 
- scaled <- srt@assays$RNA$data %>%
-   add_cols(-cell_means) %>%
-   multiply_cols(1 / cell_vars) 
- mat <- write_matrix_dir(scaled[VariableFeatures(srt),], tempfile("mat")) 
- svd <- BPCells::svds(mat, k=100, threads = 6) 
- pca <- multiply_cols(svd$v, svd$d)

`

bnprks:
Hi @Sandman-1, thanks for your question! First off, 40 hours is much longer than this operation should take -- I think 4 million cells with 2k variable features and k=50 should take under 30 minutes single-threaded, and probably under 10 minutes with 6 cores (though this can vary a bit depending on sequencing depth and choice of variable gene count).

One issue I see is that you should call `write_matrix_dir` before you call `add_cols`. The reason is that `add_cols` will cause your matrix to no longer be sparse (almost no entries will be equal to exactly 0). Although BPCells can handle `add_cols` efficiently during `svds`, the `write_matrix_dir` call loses that information.

Another slightly strange thing in your above code is that normally people scale by the gene mean and variance, not the cell stats. This may be changing your matrix in a way that causes the decomposition to fail. (From searching the TridiagEigen error, it comes from RSpectra / Spectra, and seems to be related to the contents and eigenvectors of specific matrices)

I'd suggest something like:
```r

srt <- FindVariableFeatures(srt, nfeatures = round(0.25*nrow(srt)))
[...]

PR #147, opened 2024-10-25
[r] Adjust default parameters for `cluster_graph_leiden()`

bnprks:
Currently, `cluster_graph_leiden()` by default will output a number of clusters that scales approximately linearly with number of cells if the `resolution` parameter is held constant. This is generally not good and leads to problems like [this](https://github.com/bnprks/BPCells/issues/125#issuecomment-2359162884) where people get thousands of clusters called on large datasets.

This pull request does the following:
- Changes default `objective_function` from `CPM` to `modularity` sets default `resolution` back to 1.
- Moves clustering-related tests to a new file and makes a basic test to confirm that the clustering functions at least don't crash. I don't know of a good way of validating the clustering is working, so not crashing seems good enough for now.

Here is the benchmarking data to justify this change. Note that Leiden modularity with resolution = 1 gives consistent cluster sizes just like Louvain, but Leiden CPM will give out a ton of clusters for large datasets unless the resolution parameter is adjusted down for large datasets.

resolution_plot

[cluster-resolution.csv](https://github.com/user-attachments/files/17516416/cluster-resolution.csv)

<details>
<summary> Click for plotting code </summary>

```r
data |>
  mutate(alg=case_match(alg, "leiden" ~ "Leiden CPM", "leiden-modularity" ~ "Leiden Modularity", "louvain" ~ "Louvain"),
         resolution=factor(as.numeric(resolution), sort(unique(as.numeric(resolution))))) |>
  ggplot(aes(cells, clusts, color=resolution)) +
  geom_line() +
  geom_point() +
  scale_x_continuous(transform="log10", guide=guide_axis_logticks(), labels=scales::label_log(), breaks=c(1e5, 1e6)) +
  scale_y_continuous(transform="log10", guide=guide_axis_logticks()) +
  scale_color_manual(values=RColorBrewer::brewer.pal(9, "BuPu")[3:9]) +
  facet_wrap("alg") +
  theme_bw() + 
  coord_fixed() +
  labs(title="Cluster counts by resolution", y="Cluster count", x="Dataset size (cells)")
```
</details>

### DIFF
--- a/r/NEWS.md
+++ b/r/NEWS.md
@@ -28,6 +28,7 @@ Contributions welcome :)
 - `trackplot_combine()` now has smarter layout logic for margins, as well as detecting when plots are being combined that cover different genomic regions. (pull request #116)
 - `select_cells()` and `select_chromosomes()` now also allow using a logical mask for selection. (pull request #117)
 - BPCells installation can now also be configured by setting the `LDFLAGS` or `CFLAGS` as environment variables in addition to setting them in `~/.R/Makevars` (pull request #124)
+- `cluster_graph_leiden()` now has better defaults that produce reasonable cluster counts regardless of dataset size. (pull request #147) 
 
 ## Bug-fixes
 - Fixed error message when a matrix is too large to be converted to dgCMatrix. (Thanks to @RookieA1 for reporting issue #95)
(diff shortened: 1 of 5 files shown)


immanuelazn:
A beautiful graph and test!
Changes all are agreeable.  Do you have any opinions on Seurat, with how they are implementing leiden clustering?  They default on using `RBConfigurationVertexPartition` rather than modularity.  This isn't directly available through `igraph`, but it is through the `leiden` python package

bnprks:
From the docs, it looks like [RBConfigurationVertexPartition](https://leidenalg.readthedocs.io/en/stable/reference.html#rbconfigurationvertexpartition) is basically the same objective function as modularity just with some constant scaling, so I believe this is consistent with the approach of Seurat

Issue #149, opened 2024-10-30
Potential "awk" bug 


akhst7:
Hi, 

I tried to update BPCells to 0.2 and run into a few compilation error and a following is the first error;

```
+ echo 'Failure running either '\''h5cc -show'\'' or: '\''h5cc -showconfig | awk -F: '\''/FLAGS|Extra libraries:/ {printf(%s' ', $2)}'\'''
Failure running either 'h5cc -show' or: 'h5cc -showconfig | awk -F: '/FLAGS|Extra libraries:/ {printf(%s , $2)}'
```
 This awk one liner, `h5cc -showconfig | awk -F: '/FLAGS|Extra libraries:/ {printf(%s , $2)}` gives an following error;
```
awk: syntax error at source line 1
 context is
	 >>> {printf(% <<< s, $2)}
awk: illegal statement at source line 1
```
I think this should be  ```h5cc -showconfig | awk -F: '/FLAGS|Extra libraries:/ {printf(```**"%s/n"** ```, $2)}```.  

My system is as follows;

```
                      'c.          
                 ,xNMM.          -----------------------------
               .OMMMMo           OS: macOS 15.1 24B83 arm64
               OMMM0,            Host: Mac13,2
     .;loddo:' loolloddol;.      Kernel: 24.1.0
   cKMMMMMMMMMMNWMMMMMMMMMM0:    Uptime: 13 hours, 26 mins
 .KMMMMMMMMMMMMMMMMMMMMMMMWd.    Packages: 245 (brew)
 XMMMMMMMMMMMMMMMMMMMMMMMX.      Shell: zsh 5.9
;MMMMMMMMMMMMMMMMMMMMMMMM:       Resolution: 3440x1440
:MMMMMMMMMMMMMMMMMMMMMMMM:       DE: Aqua
.MMMMMMMMMMMMMMMMMMMMMMMMX.      WM: Quartz Compositor
 kMMMMMMMMMMMMMMMMMMMMMMMMWd.    WM Theme: Blue (Dark)
 .XMMMMMMMMMMMMMMMMMMMMMMMMMMk   Terminal: iTerm2
  .XMMMMMMMMMMMMMMMMMMMMMMMMK.   Terminal Font: Monaco 10
    kMMMMMMMMMMMMMMMMMMMMMMd     CPU: Apple M1 Ultra
     ;KMMMMMMMWXXWMMMMMMMk.      GPU: Apple M1 Ultra
[...]

PR #150, opened 2024-10-30
[r] Fix awk typo in configure error message

bnprks:
This is to address some confusion that popped up in #149, where the error message didn't print out the quotation marks from the original command correctly


immanuelazn (approved):
Looks good!

PR #162, opened 2024-11-18
Ia/matrix projections lsi

immanuelazn:
## Details 
This PR allows for projection into LSI spaces, as planned with an sklearn-esque interface.  We port over both variable feature selection, and LSI.  `fit()`, `project()` and `short_description()` methods have been created for both these steps. 

Currently the implementations utilize a lot of the previously created LSI implmementation.  I am refraining from moving that over until review is concluded.  However, this PR already contains some initial tests, and a demonstration that we can concat two `PipelineSteps` together, to fit and project.

## Tests
- Tests were created to make sure initialization, as well as `fit()` and `project()` methods work for variable feature selection and LSI.
- Tests were also created to make sure pipelines can be created between these two individual steps, with `fit()` and `project()` working from the `Pipeline` implementations of these two methods


immanuelazn:
Closed as we are changing direction to follow #167

PR #163, opened 2024-11-25
[cpp] update hwy to v1.2.0

immanuelazn:
## Details

The new highway updates allow for templating of functions, which could potentially allow us to de-dupe some of our math code that is written with multiple types (see `cpp/simd/math*`)

Writing code with templates can be done following the example in **[`hwy/highway.h`](https://github.com/google/highway/blob/2b565e87d50b151660494624af532ac0b6076c79/hwy/highway.h#L400)**


bnprks:
Yeah, it seems like a good idea to update since `hwy` is making steady improvements. I think there are two bits I'd like to nail down:

1. Precisely which commit form the highway repo is this updating to? From running some diffs it doesn't appear to be the commit tagged 1.2.0 in highway's repo. We don't necessarily need a tagged version, but do need at least a commit hash so it's clear exactly where the code came from.
2. I'd like to make sure the updates to `build_highway.sh` are informed by a diff of highway's CMakeLists.txt between 1.0.5 and this version.

    - The goal is to try to keep build_highway.sh maintaining some correspondence to the latest CMakeLists.txt so we are adding the same compiler flags Highway does.
    - We can look at this together if you like. I don't think there are that many changes to consider but it's always challenging to parse through a big cmake build file

PR #168, opened 2024-12-12
[r] add tf-idf and log normalization functions

immanuelazn:
# Details
As discussed, we are looking to add in normalization functions to allow for passing in transformations into orchestrator functions like LSI/iterative LSI.  I add in two functions, `normalize_tfidf()`, and `normalize_log()` (shown in #167).  

There were a few departures from the design doc, in order to provide a little bit more flexibility.  Particularly, I was thinking about the case where the feature means are not ordered in the same way.  To add in a little bit of safety, I added some logic for index invariant matching for feature means to matrix features.

Other than that, I also provided an option to do a traditional log transform by boolean flag, rather than log1p.  As we don't directly expose a log function in BPCells C++ side, I just added a - 1.  However, I'm noticing that this isn't translated into a -Inf like in dgCMatrix/generic matrix, and is instead a very small number. Might need to evaluate if this is something we would want to support


r/tests/testthat/test-matrix_transforms.R
```diff
@@ -346,3 +346,49 @@ test_that("linear regression works", {
...
+    res_1 <- as(normalize_log(m2), "dgCMatrix")
+    expect_equal(res_1, log1p(m*1e4), tolerance = 1e-6)
+    
+    # Test that changing scale factor works
+    res_2 <- as(normalize_log(m2, scale_factor = 1e5), "dgCMatrix")
+    expect_equal(res_2, log1p(m*1e5), tolerance = 1e-6)
+    # Test that removing the add_one works
+    # log of 0 is -inf, but we don't do that on the c side, and just have really large negative numbers.
+    res_3 <- as(normalize_log(m2, add_one = FALSE), "dgCMatrix")
+    res_3@x[res_3@x < -60] <- -Inf
```
immanuelazn, line 392: Any better way of doing this?

r/R/transforms.R
```diff
@@ -923,3 +923,57 @@ regress_out <- function(mat, latent_data, prediction_axis = c("row", "col")) {
...
+
+
+#' Normalize a `(features x cells)`` matrix using term frequency-inverse document frequency
+#' @param mat (IterableMatrix) to normalize
+#' @param feature_means (numeric) Means of the features to normalize by.  If no names are provided, then 
+#' each numeric value is assumed to correspond to the feature mean for the corresponding row of the matrix.
+#' Else, map each feature name to its mean value.
+#' @returns tf-idf normalized matrix.
+#' @export
+normalize_tfidf <- function(mat, feature_means = NULL, threads = 1L) {
```
bnprks, line 955: - Nice touch thinking of the name-matching option for `feature_means`
- I think we might want to follow the ArchR/Signac TF-IDF default formulas which include some logarithms. Perhaps you were thinking users would combine `normalize_log` with `normalize_tfidf` but I think it's better to keep each standalone
- We should also give the specific normalization formula here

r/tests/testthat/test-matrix_transforms.R
bnprks: As suggested above, I think we just get rid of the `add_one` option

bnprks (commented):
This overall looks reasonable, though I'm not sure if the formulas implemented are quite what we want.

For documenting the formulas, I'd suggest taking advantage of the fact that we can render latex inside `\eqn{}` blocks from docstrings. Due to a current [pkgdown bug](https://github.com/r-lib/pkgdown/issues/2704) we'll need to edit `_pkgdown.yml` to un-break our equation rendering:

```yaml
includes:
    in_header: |
       <script defer data-domain="benparks.net" src="https://plausible.benparks.net/js/visit-counts.js"></script>
       <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.15/dist/katex.min.css" integrity="sha384-Htz9HMhiwV8GuQ28Xr9pEs1B4qJiYu/nYLLwlDklR53QibDfmQzi7rYxXhMH/5/u" crossorigin="anonymous">
       <!-- The loading of KaTeX is deferred to speed up page rendering -->
       <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.15/dist/katex.min.js" integrity="sha384-bxmi2jLGCvnsEqMuYLKE/KsVCxV3PqmKeK6Y6+lmNXBry6+luFkEOsmp5vD9I/7+" crossorigin="anonymous"></script>
       <!-- To automatically render math in text elements, include the auto-render extension: -->
       <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.15/dist/contrib/auto-render.min.js" integrity="sha384-hCXGrW6PitJEwbkoStFjeJxv+fSOOQKOPbJxSfM6G5sWZjAyWhXiTIIAmQqnlLlh" crossorigin="anonymous" onload="renderMathInElement(document.body);"></script>
```

I'd suggest documenting the transformation equation for a single matrix element, like this for tfidf: `eqn{\tilde{x}_{ij} = \log(\frac{x_{ij} \cdot \text{scaleFactor}}{  \text{rowMean}_i\cdot \text{colSum}_j} + 1)}`

This would render as:
image

r/R/transforms.R
```diff
@@ -923,3 +923,57 @@ regress_out <- function(mat, latent_data, prediction_axis = c("row", "col")) {
...
+# Normalizations
+#################
+
+#' Normalize a matrix using log normalization
+#' @param mat (IterableMatrix) Matrix to normalize
+#' @param scale_factor (numeric) Scale factor to multiply matrix by for log normalization
+#' @param add_one (logical) Add one to the matrix before log normalization
+#' @returns log normalized matrix.
+#' @export
+normalize_log <- function(mat, scale_factor = 1e4, add_one = TRUE) {
```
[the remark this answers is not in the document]
immanuelazn: Agree on removing the add_one, and providing formulas in docs

immanuelazn:
Thanks for the review Ben!  I agree with the styling changes.  I believe a small discussion on the logic of these functions would be a good idea to make sure we're aligned.

My initial thought was that `normalize_log()` and `normalize_tfidf()` would be to the tee of what these functions are named as.  In the case of log normalization we would have the division by cell sums as built in the wrapper function (such as in `select_features_by_variance()` instead).  What I'm worried about is in cases where we would want to pass the function, but we don't want a separate cell sum normalization as well.  `select_features_by_dispersion()` is an example in the design docs, which has an optional log normalization _without doing cell normalization_  This reflects the design decisions in ArchR [here](https://github.com/GreenleafLab/ArchR/blob/d9e741c980c7c64e5348c97a74d146cc95f8ba76/R/IterativeLSI.R#L1023).

Similarly, from what I recall, tf-idf normalization doesn't necessarily indicate that a log normalization will take place after it.  I was thinking in the case of LSI/iterative LSI, we would again just have the log normalization as a required step post tf-idf normalization.    
 

I think having a boolean flag for cell normalization for `normalize_log()`, as well as having a boolean flag for log normalization in `normalize_tfidf()` would provide a little bit more flexibility.  What do you think?

bnprks:
I'm thinking of `normalize_log()` and `normalize_tfidf()` as trying to simplify single cell workflows by implementing common normalizations. We don't need these to cater to every possible need, just focus on making common cases easy (hard cases can be implemented by hand using our existing math operators). I think the versions I've suggested are the common cases

- I think the ArchR example you link is actually an example of ArchR performing no normalization rather than doing a log without normalizing by read depth. That's a use-case we can easily support by allowing `normalize=NULL` or something to that effect.
- I don't think I've ever seen a case of log normalization not include normalizing by colSums first. (e.g. this is what [Seurat](https://github.com/satijalab/seurat/blob/1549dcb3075eaeac01c925c4b4bb73c73450fc50/R/preprocessing5.R#L247) does, when it says "logNormalize"). I really don't think people will be confused here, and we can include the formula in the docs for clarity. I don't think we need a boolean flag unless we have a known example of a common analysis approach it would enable.
- For tfidf, it would be okay to implement `method` argument from [Signac](https://github.com/stuart-lab/signac/blob/0b33b1154f9a610897d1efad2c0065081c4e7132/R/preprocessing.R#L646), though just providing the default method is okay, as probably 99% of people just use the default that both ArchR/Signac use. Annoyingly Signac and ArchR disagree on `method` and `LSIMethod` numbering scheme, so we might need to use string naming if we implement this. (again, optional for now to implement the other variants. Implementing the ArchR/Signac default is most important)

immanuelazn:
There was a previous decision to try to split PRs into more bite-sized pieces.  We decided that this was going too far in this direction, as there becomes many layers of PRs stacked on top of each other.   As a result, this PR and the feature selection PR will both point to main.

Issue #173, opened 2024-12-20
Weird pseudobulk result with fractions after thresholding


jonathan-columbiau:
Hi, thanks for creating the BPCells package! I found a weird result that I think is a bug while performing an analysis where I am trying to get the fraction of cells per group that express each gene above a certain CPM threshold, while using the pseudobulk_matrix function. Ex: the max value for one of the columns is 4.645053e+296, when it should be between 0 and 1.

Here's some code that replicates the error using some of the code in the introductory vignette: 

library(BPCells)
suppressPackageStartupMessages({
    library(dplyr)
})

data_dir <- file.path(tempdir(), "pbmc-3k")
dir.create(data_dir, recursive = TRUE, showWarnings = FALSE)
setwd(data_dir)

if (!file.exists("pbmc_3k_rna_raw")) {
    mat_raw <- open_matrix_10x_hdf5("pbmc_3k_10x.h5", feature_type="Gene Expression") %>% 
        write_matrix_dir("pbmc_3k_rna_raw")
} else {
    mat_raw <- open_matrix_dir("pbmc_3k_rna_raw")
}
keeper_cells <- colnames(mat_raw)[Matrix::colSums(mat_raw) > 1e3]

keeper_genes <- Matrix::rowSums(mat_raw) > 3
mat <- mat_raw[keeper_genes,keeper_cells]

mat <- multiply_cols(mat, 1/Matrix::colSums(mat) * 1e6)

mat_greater_3_cpm = mat > 3

#pseudobulk by mean to get fraction of genes greater 3 cpm 

groups = c(rep("A",1000),rep("B",1000),rep("C",724)) %>% as.factor()

temp_mat = write_matrix_dir(mat_greater_3_cpm,dir = tempfile("mat2"))
fraction_cells_per_gene_greater_3_cpm = pseudobulk_matrix(temp_mat,groups, method = "mean")

bnprks:
Hi @jonathan-columbiau, thanks for the bug report with the reproducible example! This is a type confusion bug in `pseudobulk_matrix` where we forgot to convert all input matrices to `"double"` type, resulting in errors when an integer matrix (like those from `mat > 3`) is produced.

I've opened up a PR to fix this, so your options are:
1. Just run `convert_matrix_type(temp_mat, "double")` before calling `pseudobulk_matrix` as a temporary workaround
2. Re-install from the fixed branch `remotes::install_github("bnprks/BPCells/r", ref="bp/173-fix")`
[...]

PR #174, opened 2024-12-20
[r] Fix type confusion in `pseudobulk_matrix()` and clean up `parallel_split()`

bnprks:
As reported in #173, `pseudobulk_matrix()` assumes that all input matrices are type "double", which results in type confusion in some circumstances. While fixing this, I also noticed an issue in `pseudobulk_matrix`'s use of `parallel_split`, which would needlessly split up a matrix even if no parallelism was requested (slight potential performance/memory hit). I did a small refactor to make `parallel_split` easier to use correctly and resolve the issue in `pseudobulk_matrix`

Changes in this PR:
- Always cast inputs to `pseudobulk_matrix()` to type `double`
- Adjust `parallel_split` logic to not do any splitting if `threads <= 1`
- Adjust other usages of `parallel_split` to take advantage of not needing the conditional logic anymore

### DIFF
--- a/r/R/singlecell_utils.R
+++ b/r/R/singlecell_utils.R
@@ -88,7 +88,7 @@ marker_features <- function(mat, groups, method="wilcoxon") {
 #' extra calculation time, and when calculating `mean`, adding `nonzeros` will take no extra time.
 #' @inheritParams marker_features
 #' @export
-pseudobulk_matrix <- function(mat, cell_groups, method = "sum", threads = 1L) {
+pseudobulk_matrix <- function(mat, cell_groups, method = "sum", threads = 0L) {
   assert_is(mat, "IterableMatrix")
   assert_is(cell_groups, c("factor", "character", "numeric"))
   assert_true(length(cell_groups) == ncol(mat))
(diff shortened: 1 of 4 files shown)


r/R/singlecell_utils.R
```diff
@@ -101,9 +101,13 @@ pseudobulk_matrix <- function(mat, cell_groups, method = "sum", threads = 1L) {
     }
   }
   assert_is(threads, "integer")
-  # if multiple methods are provided, only need to pass in the top method as it will also calculate the less complex stats
-  iter <- iterate_matrix(parallel_split(mat, threads, threads*4))
-  res <- pseudobulk_matrix_cpp(iter, cell_groups = as.integer(cell_groups) - 1, method = method, transpose = mat@transpose)
+
+  it <- mat %>%
+    convert_matrix_type("double") %>%
+    parallel_split(threads, threads*4) %>%
```
immanuelazn, line 107: Think this is reasonable to use the piping style instead.  

From what I understand, this forces all matrices no matter what type to be converted to doubles and sent to `pseudobulk_matrix<double>`. While this completely works, I question whether it makes sense to require usage of `pseudobulk_matrix<double>` as we have explicit instantation for `float` and `uint32_t`.  

I wonder if it makes sense to just properly handle this at the `matrix_utils.cpp` file level on `pseudobulk_matrix_cpp()`. Rather than forcing everything to work on matrices of type `double`, could we do something like how `SEXP iterate_matrix_col_select_<type>_cpp()` handles this, by using a different function depending on type, and doing a `do.call()` to perform a string method search for the correct function?  Either that, or passing a `type` param in to `pseudobulk_matrix_cpp()`, to determine which `pseudobulk_matrix` cpp type to call.

immanuelazn (approved):
Looks good! I enclosed a general question regarding how we want to do our typing going fowards though

r/R/singlecell_utils.R
bnprks: Adding in proper type handling would be a great option. Right now several parts of BPCells do conversions to doubles for math, which is not terrible for efficiency when most of the calculations return doubles anyhow, but there's definitely room for improvement.

Properly utilizing the non-double implementations would be an interesting bit of follow-up work. With the plans for the C++/R/python refactor it might largely solve itself, but it's also not too hard to fix up in the current code base.

r/R/singlecell_utils.R
```diff
@@ -112,7 +116,8 @@ pseudobulk_matrix <- function(mat, cell_groups, method = "sum", threads = 1L) {
   }
   # give colnames and rownames for each matrix in res, which is a named list
   for (res_slot in names(res)) {
-    if ((length(res[[res_slot]]) == 0) || !(res_slot %in% method)) {
```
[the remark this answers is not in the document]
bnprks: Yeah, it only really affects matrices where one dimension is zero. Most of the time doesn't matter but I think Seurat in the past has surfaced bugs of BPCells not handling zero-dimensions consistently so it's nice to make them work when we notice

PR #179, opened 2024-12-23
[r] Fix perf subsetting large concatenated matrices

bnprks:
The logic for subsetting rbind or cbind matrices was taking time O(length(selection)* # sub-matrices), which could be quite slow for large datasets. Now, rather than performing a linear search through the selection indices to find the indices in range for each sub-matrix, we just do a binary search which eliminates the performance issue.


immanuelazn (approved):
The new local_i, and local_j search is really clever.  Definitely was faster on my end, and only some minor copy/paste changes.

As a small thought exercise though, I think the main problem was running `local_i <- i$subset[i$subset >= row_start & i$subset <= row_end] - last_row` repetitively. We can notice that `split_Selection_index()` sorts indices already. I think this could have been replaced with a two pointer approach with the cumulative sums against `i$subset`, which would only require one pass through `i$subset`.  

Reading through the [`findInterval`](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/findInterval) documentation though, since the first arg is sorted by definition, it can be `O(length(cumsum(c(0, rows)))` even though a binary search is used.  Overall the differences in speed shouldn't be substantial though!

bnprks:
The findInterval logic is definitely a little trickier to follow than I'd like. If we were in C++ I'd definitely go for the two pointer approach, but I'm not actually sure if R has a good way to search arrays like that since R for loops are pretty suspect from a performance perspective.

Thanks for spotting those copy-paste editing issues!

PR #183, opened 2025-01-10
[r] Use whole number check instead of integer check for `pseudobulk_matrix()`

immanuelazn:
Self explanatory, writing `L` after a number feels unnecessary.


bnprks:
Looks good, and seems like elsewhere in the code base we're using `assert_is_wholenumber` already

PR #190, opened 2025-01-27
[r] provide more mat input flex to `matrix_stats()`

immanuelazn:
Pretty self explanatory, added a few lines to allow matrix_stats, to be a little more input matrix agnostic.  I thought it would be better for us to automatically try to convert with a try, except block.  This should allow downstream caller functions to also work with `matrix` and `dgcMatrix`.

Also added a small helper conversion between `matrix` and `IterableMatrix`.


bnprks:
I just made a couple tweaks in here -- I checked the documentation for `as()` and found out about the `canCoerce` function which seems a bit cleaner, and also made the error message go through `rlang::abort` which I think is preferable for most new error messages we write due to the nice formatting.

I'm happy with all the changes now, so if my tweaks look fine to you then go ahead and to the squash + merge

PR #191, opened 2025-01-27
[r] import matrix transpose generic

immanuelazn:
Stemming from a discussion back in October 24, BPCells transpose breaks when importing a package that uses `BiocGenerics`  (such as ArchR and many Bioconductor packages).  The following error would occur even when input is an `IterableMatrix`: 
```Error in t.default(mat): argument is not a matrix```  

Interestingly enough, loading BPCells prior to loading `BiocGenerics` worked, but the converse didn't.  We follow guidance from (https://github.com/r-lib/pkgload/issues/88), to prevent needing to create our own generic for `t()`.


bnprks (approved):

bnprks:
To clarify, it seems this issue only happened when BPCells was loaded with `devtools::load_all()`, not when it was loaded via `library(BPCells)`. Still, it doesn't seem to hurt and makes development a bit easier.

PR #192, opened 2025-01-27
[r] fix gene score params

### LINKED ISSUE #185: tile_width only works with 500bp
In `gene_score_weights_archr`, the tile_width is not transferred to `gene_score_tiles_archr` which lack the `tile_width = tile_width` argument.

https://github.com/bnprks/BPCells/blob/main/r/R/geneScores.R#L206

### LINKED ISSUE #188: Raise error when chromosome_sizes not sorted
I found out that if `chromosome_sizes` is not sorted when using `gene_score_archr` the gene score matrix output is wrong.

This is due to the `cumsum` of the tiles in `gene_score_weights_archr` when trying to find the starting tile of each chromosome.

https://github.com/bnprks/BPCells/blob/main/r/R/geneScores.R#L219C1-L219C48

I think it would be beneficial for the user to raise more than a warning `order_ranges()` as the outcome would be completely inaccurate.

immanuelazn:
Fixes #185

Params for tile width in `gene_score_width_archr()` weren't being passed down previously.


bnprks:
I added a fix for the `chromosome_sizes` ordering, along with some regression tests to help improve our test coverage while we're at it. Ready to merge from my perspective

PR #200, opened 2025-02-04
[r] fix unintended removal of NEWS line from previous PR

immanuelazn:
Not entirely sure why merging from main in #134 removed a line in NEWS, but this reverts that removal


bnprks:
I did a quick edit to put the line back in the original location rather than shifted by one in the order

PR #206, opened 2025-03-08
[r] clean up svds docs

immanuelazn:
I've noticed that the math doesn't get rendered correctly in `Details`. Looks like a copy-pasta from `RSpectra`, but `\eqn` blocks in `\describe{\item{}}` don't render correctly in our book.  I did a minimal fix by converting the offending lines into code lines instead.

See the offending reference [here](https://bnprks.github.io/BPCells/reference/svds.html).

Also fixed numbers and variables to be repesented as code blocks across params.

### DIFF
--- a/r/R/matrix-svds-docs.R
+++ b/r/R/matrix-svds-docs.R
@@ -17,8 +17,7 @@
 #'
 #' @param A The matrix whose truncated SVD is to be computed.
 #' @param k Number of singular values requested.
-#' @param nu Number of left singular vectors to be computed. This must be between 0 and 'k'. (Must be equal to 'k' for BPCells IterableMatrix) 
-#' @param nu Number of right singular vectors to be computed. This must be between 0 and 'k'. (Must be equal to 'k' for BPCells IterableMatrix) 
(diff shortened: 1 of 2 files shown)


bnprks:
Nice initiative on putting in a suggested fix, but I think the equation rendering is fixed already by a pending change that's currently in PR 189, as described in my [comment](https://github.com/bnprks/BPCells/pull/168#pullrequestreview-2501053329) on PR 168. It appears the pkgdown bug has a fix that was merged about a week ago, but it'll still be some time before it gets into a CRAN release I think.

It took me a while to re-figure out what was going on, because in the current main branch with a latest pkgdown installed it's impossible to get *any* equation to render correctly. I think with the fix I linked applied all of the equations render just fine (though some of your other changes like getting rid of the duplicate parameter comment seem good still).

Also, I think for things that are mostly fixing typos in docs/error messages/warning messages I think it's okay not to list as a separate item in the NEWS file

immanuelazn:
Cool, thanks!  I forgot about that change, was mainly concerned about how the `\eqn` blocks that were working in my #189  weren't working when copy pasted into this branch's lists.  But probably should have done a due diligence check in params too.  

In any case, I'll remove those changes and just keep the docs styling changes for the non equation blocks.

PR #214, opened 2025-03-24
[r][cpp] Force pow to use strictly positive exponent

bnprks:
I realized that we shouldn't support negative exponents for powers, since that would mess up all the zero values in the matrix. Just updated a few bounds checks to make that happen


immanuelazn (approved):
Looks good.  Should we consider adding this info to the `IterableMatrix Methods` reference?

PR #233, opened 2025-04-09
[r] Add examples for single cell plots and trackplots

immanuelazn:
This PR provides function examples for most of the functions in the `Single Cell plots` and `Genomic track plots` section.  
This PR also provides a function (`render_plot_from_storage()`) that allows for resizing plots in an example, as pkgdown does not natively provide configuration on that end.

Missing functions include the following:

- `plot_tf_footprint()`
  - This requires a lot of setup infra and I don't see an easy way of generating fake data that makes sense for these footprints.  Should I just go ahead and add in the data preperation code anyways?
- `set_trackplot_label()` `set_trackplot_height()` and `get_trackplot_height()`.  
  - I think all of these functions are dependent on the `trackplot_combine()` function.  This would turn into an entire mini-vignette with customizing the `trackplot_combine()` plot.  Should this be the goal?

There are some various thoughts, and issues with the plots, that I will note directly in code. However, there is a large issue that likely comes from `argument_name()` that causes the plots in pkgdown to be different than what is rendered in console.  I tried with and without rendering to storage, but the issue still persists.


r/R/plots.R
```diff
@@ -485,6 +520,48 @@ collect_features <- function(source, features = NULL, gene_mapping = human_gene_
 #' @return By default, returns a ggplot2 object with all the requested features plotted
 #' in a grid. If `return_data` or `return_plot_list` is called, the return value will
 #' match that argument.
+#' @examples
```
immanuelazn, line 523: I'm rendering two plots, once by just the clusters, and one with the features.  Not really understanding why this is erroring out, when running this code in console runs completely fine.

r/R/plots.R
```diff
@@ -485,6 +520,48 @@ collect_features <- function(source, features = NULL, gene_mapping = human_gene_
...
+## Prep data
+#' set.seed(123)
+#' mat <- get_demo_mat()
+#' ## Normalize matrix
+#' mat_norm <- log1p(multiply_cols(mat, 1/colSums(mat)) * 10000) %>% write_matrix_memory(compress = FALSE)
+#' ## Get variable genes
+#' stats <- matrix_stats(mat, row_stats = "variance")
+#' variable_genes <- order(stats$row_stats["variance",], decreasing=TRUE) %>%
```
immanuelazn, line 531: Not really liking doing all of this setup where I'm doing PCA and clustering.  I think this can be vastly shortened when we have the #189 feature added in + a PCA `DimReduction`.  But should we just do it with a fake matrix created by manually concatting multiple different distributions?

r/R/plots.R
```diff
@@ -741,6 +818,17 @@ rotate_x_labels <- function(degrees = 45) {
 #' @inheritParams plot_embedding
 #' @param gene_mapping An optional vector for gene name matching with match_gene_symbol().
 #' @param colors Color scale for plot
+#' @examples
+#' ## Prep data
+#' mat <- get_demo_mat()
+#' cell_types <- paste("Group", rep(1:3, length.out = length(colnames(mat))))
+#' 
+#' ## Plot dot
```
immanuelazn, line 826: Same issue with the y label where it's not being rendered properly.  I think it is again with the `argument_name()` navigating up the call stack.

r/R/trackplots.R
```diff
@@ -510,6 +579,18 @@ trackplot_coverage <- function(fragments, region, groups,
 #' @param label_size size for transcript labels in units of mm
 #' @return Plot of gene locations
 #' @seealso `trackplot_combine()`, `trackplot_coverage()`, `trackplot_loop()`, `trackplot_scalebar()`
+#' @examples
```
immanuelazn, line 582: Side thought but maybe it would be good to split into to y-axis levels even when there isn't direct intersection of two genes.  It is a little bit hard to distinguish the genes in this example.

r/R/trackplots.R
```diff
@@ -268,6 +284,39 @@ trackplot_normalize_ranges_with_metadata <- function(data, metadata) {
 #'    the text label, y-axis, and plot body. The relative height of each row is given
 #'    by heights. A shared title and x-axis are put at the top.
 #' @seealso `trackplot_coverage()`, `trackplot_gene()`, `trackplot_loop()`, `trackplot_scalebar()`
+#' @examples
```
immanuelazn, line 287: Should we expand the trackplot_combine() section, or is this good for now?

r/R/trackplots.R
```diff
@@ -607,6 +688,28 @@ trackplot_gene <- function(transcripts, region, exon_size = 2.5, gene_size = 0.5
 #' @param show_strand If TRUE, show strand direction as arrows
 #' @return Plot of genomic loci if return_data is FALSE, otherwise returns the data frame used to generate the plot
 #' @seealso `trackplot_combine()`, `trackplot_coverage()`, `trackplot_loop()`, `trackplot_scalebar()`, `trackplot_gene()`
+#' @examples
```
immanuelazn, line 691: The peaks here don't look great, should I use a different example?

PR #240, opened 2025-04-14
[r] Fix `libhwy` detection

### LINKED ISSUE #228: Faulty `libhwy` detection in R package
The code in the R package for detecting an existing `libhwy` appears to always fail and use the vendored copy. Specifically, flags are not wired into the test:

https://github.com/bnprks/BPCells/blob/9d848ca44ca01119d96bf5e9706c4074da41e6ee/r/configure#L158-L162

~~Also, note the value in `HWY_CFLAGS` doesn't appear necessary.~~ (Nevermind, it is indeed required.)

While visiting this code, we also observe an incorrect variable being tested (should be `HWY_OK`):

https://github.com/bnprks/BPCells/blob/9d848ca44ca01119d96bf5e9706c4074da41e6ee/r/configure#L173-L176

---

I am already patching this [for a Bioconda build](https://github.com/bioconda/bioconda-recipes/pull/55118). I'd be happy to send along a PR with suggested adjustments once I have that going.

mfansler:
Resolves #228 

Conditional logic for messages is also slightly adjusted.

### DIFF
--- a/r/configure
+++ b/r/configure
@@ -158,21 +158,23 @@ printf "\nTesting availability of highway SIMD library..."
 HWY_OK=""
 HWY_CFLAGS="-Ibpcells-cpp"
 HWY_LIBS="-lhwy"
-$CXX tools/hwy-test.cpp $CXXFLAGS $LDFLAGS 2>&3 && HWY_OK="yes";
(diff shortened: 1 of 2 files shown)


bnprks:
Thanks for the PR @mfansler! Looks like a nice fix + cleanup of the build logic

PR #256, opened 2025-05-09
Use `int32_t` and `int64_t` for 10x H5 writing function

ycli1995:
Hi, @bnprks 
When I tried to use `BPCells::write_matrix_10x_hdf5` in `loupeR` from https://github.com/10XGenomics/loupeR , the program `louper.exe create --input=<h5file> --output=<cloupe file>` raised an error:
```
indices: cannot convert dataset to int64
```
When I change the original `uint32/64_t` to `int32/64_t` in "indices" and "indptr", the error is fixed. Maybe it would be better if the HDF5 files written by `BPCells` use data types more consistent with 10x's expected formats, to improve compatibility with downstream tools.

### DIFF
--- a/r/NEWS.md
+++ b/r/NEWS.md
@@ -10,7 +10,8 @@
 - Fixed memory errors when running `writeInsertionBed()` and `writeInsertionBedGraph()` (pull request #{118, 134})
 - Export `merge_peaks_iterative()`, which helps create non-overlapping peak sets.  (pull request #216)
 - Add support for `uint16_t` when reading in anndata matrices using `open_matrix_anndata_hdf5()`. (pull request #248)
-
+- Switch `write_matrix_10x_hdf5()` to use signed rather than unsigned integers for `indices`, `indptr`, and `shape` to improve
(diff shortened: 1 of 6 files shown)


bnprks:
Hi @ycli1995, thanks for finding + fixing this!

One small note I observed -- 10x files seem to use int64_t for the "indices" field, not "int32_t". (e.g. downloading from [this public dataset](https://www.10xgenomics.com/datasets/2500_Wistar_Rat_PBMCs_Singleplex_3p_gem-x_Universal))

I've adjusted the "indices" type with the commits I just pushed, as well as adding a small test and an update to the change log. If that all looks good to you then let me know and I'll merge in -- otherwise happy to discuss more and revert any of my changes if needed.

ycli1995:
Thanks for the quick follow-up!

Using `int64_t` looks good to me. Perhaps the potential downside is the increased storage cost compared to `int32_t`, but for compatibility with 10x tools like `louper`, it's definitely better to stay consistent with their format.

Feel free to merge your changes — they look great to me!

PR #265, opened 2025-06-13
[r] remove error from `cbind` and `rbind`, when matrices are of different types, to upcast instead of erroring out.

immanuelazn:


bnprks:
A couple comments:
1. I think we should unconditionally upcast to `double` to here, since there are some numbers that `uint32_t` can represent that `float` can't represent exactly (float only covers 24-bit integers exactly). I admit it's not a likely scenario, but given that we default so much math to double anyhow we might as well take the safer option.
2. To future-proof this in the event of added types, maybe we should error out if the input type isn't one of `double` `uint32_t` or `float` with something like the original error message.
3. The warning message should start with some context that lets the user know where the problem is coming from. For example, issue 261 ran into the message in the middle of some Seurat operations so the message should be interpretable even for users who aren't directly calling `rbind()` or `cbind()`

r/R/matrix.R
```diff
@@ -1109,15 +1109,15 @@ setMethod("rbind2", signature(x = "IterableMatrix", y = "IterableMatrix"), funct
...
-        matrix_type(x), matrix_type(y), new_type
+      sprintf(
+        paste0("Mismatching matrix types when calling %sbind().\n",
+        "First matrix type: %s, second matrix type: %s.\nUpcasting to double to match matrix types"), 
+        ifelse(x@transpose, "c", "r"), matrix_type(x), matrix_type(y)
       )
     )
-    x <- convert_matrix_type(x, new_type)
-    y <- convert_matrix_type(y, new_type)
+    if (matrix_type(x) != "double") x <- convert_matrix_type(x, "double")
```
bnprks, line 1119: Minor nit: you technically don't need the if statement, as `convert_matrix_type` already no-ops if the matrix is already the desired type

r/R/matrix.R
```diff
@@ -1109,15 +1109,15 @@ setMethod("rbind2", signature(x = "IterableMatrix", y = "IterableMatrix"), funct
...
   if (matrix_type(x) != matrix_type(y)) {
     # upcast if mismatching types
-    type_order <- c("uint32_t", "float", "double")
-    new_type <- type_order[max(match(matrix_type(x), type_order), match(matrix_type(y), type_order))]
     rlang::warn(
-      sprintf("First matrix type: %s, second matrix type: %s.\nUpcasting to %s to match matrix types", 
-        matrix_type(x), matrix_type(y), new_type
+      sprintf(
+        paste0("Mismatching matrix types when calling %sbind().\n",
+        "First matrix type: %s, second matrix type: %s.\nUpcasting to double to match matrix types"),
```
bnprks, line 1115: I'm realizing this warning might get printed multiple times if someone is calling `rbind` or `cbind` to merge many matrices at once. Maybe we can shorten it to one line: `"%sbind2(): Mismatching matrix types (%s vs. %s). Upcasting to double`

As a side note, `rlang::warn()` supports character vectors for its first arg in which case it makes nicely bulleted output (better than manual newlines). But in this case it's probably best to just squeeze on one line.

r/tests/testthat/test-matrix_utils.R
```diff
@@ -278,10 +278,10 @@ test_that("rbind and cbind check types (#68 regression)", {
   m <- generate_dense_matrix(10, 5) %>% as("dgCMatrix") %>% as("IterableMatrix")
 
   expect_warning(
-    rbind(m, convert_matrix_type(m, "uint32_t")), "type"
+    rbind(m, convert_matrix_type(m, "uint32_t")), "types when calling rbind"
```
bnprks, line 281: I like your logic to print the correct of `rbind` or `cbind`. Maybe also run a test case of `rbind(t(m), ...)` to test that the warning message is also correct in the transposed case?

immanuelazn: So I realized maybe this doesn't make too much sense to expose the backend logic that rbind might be calling cbind if it's transpose (or vice versa), since most people won't recognize that the backend representation of the data doesn't change with transpose status.  

I think I'll change it back to rbind always printing "rbind", cbind always printing "cbind".  But will add the test cases too though

bnprks:
Looks great to me with those revisions!

Perfectly reasonable to switch back to simple logic with the `rbind()` and `cbind()` error/warning messages. 

I think a case where someone concatenates N matrices with `double`, `uint32_t`, `uint32_t`, ... will probably result in N-1 warning messages, but I tried checking if we can directly implement `rbind` methods rather than `rbind2` and I don't think it's practically possible as they don't quite use the standard generic machinery.

immanuelazn:
Always willing to revisit a way to clean up the warnings later on!

PR #334, opened 2026-04-08
[r] Fix `knn_hnsw()` input query rowname bug

erliuu:
This pull request fixes a bug in `knn_hnsw()` where the rownames of the result matrices were incorrectly taken from `data` instead of `query` when a separate query matrix was provided. This caused the output to have wrong rownames (from the reference data) rather than the rownames corresponding to the actual query points. When no query is given, `query` is already set to `data`, so behavior is unchanged in that case. (`r/R/clustering.R`)

### DIFF
--- a/r/R/clustering.R
+++ b/r/R/clustering.R
@@ -387,8 +387,8 @@ knn_hnsw <- function(data, query = NULL, k = 10, metric = c("euclidean", "cosine
     }
   }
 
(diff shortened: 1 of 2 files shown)


bnprks:
Looks great, thanks! Sorry about the broken CI checks, but tests good locally and is pretty obviously good

Issue #296, opened 2025-08-28
Addition of 2 matrices


yxsee:
Hi Ben,
I saw that there is a function for adding a constant or vector to an IterableMatrix, but is there a function for addition of 2 IterableMatrix together? Something like below. Thanks!

1 2 3 
1 2 3
 
(+)

2 3 4 
3 4 5 

=

3 5 7 
4 6 8

immanuelazn:
Hi, we don't have anything for matrix matrix addition (that I am aware of).  but I can add this to the priority list!

erliuu:
Element-wise IterableMatrix addition (and subtraction) have been added to main with PR #336. Thanks for the suggestion @yxsee!

PR #336, opened 2026-04-24
[r][cpp] add element-wise matrix addition

erliuu:
Hi @bnprks, here are the matrix addition changes we discussed.
Resolves outstanding to-do in NEWS.md and Issue #296.

### Description
Prior to this change, combining two matrices required intermediate materialization in memory with `dgCMatrix`, resulting in high memory overhead for very large matrices. This pull request adds a lazy element-wise addition of two `IterableMatrix` objects through the existing `+` operator.
- Adds a `MatrixAddition` class that iterates column-by-column and sums via a dense accumulator mirroring `SparseMultiply`.
- Overrides `colSums`/`rowSums`/`vecMultiplyRight`/`vecMultiplyLeft` via distributive/linearity properties to bypass accumulation
- Follows `MatrixMultiply` subsetting and `rbind2`/`cbind2` `merge_dimnames()` conventions.

### Behavior
- Matrix size requirements for element-wise addition are enforced (identical dimensions)
- Type coercion: mismatched matrix types promote to `double`, equal types preserved.
- Transpose: both matrices must share internal transpose states (users are pointed to `transpose_storage_order()`)

### Tests
- Added one `MatrixMath.Add` test to align with existing arithmetic-loader conventions
- Added basic addition, type coercion, transpose, dimnames, and dim-mismatch tests

### Scope
- Element-wise subtraction is not in this PR. Users can still subtract explicitly via `mat1 + (mat2 * -1)`. To discuss if subtraction should be implemented as a separate function.

### DIFF
--- a/r/R/transforms.R
+++ b/r/R/transforms.R
@@ -752,27 +752,6 @@ setMethod("+", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
   e2 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e2, "double"))
   e2 + e1
 })
-setMethod("+", signature(e1 = "IterableMatrix", e2 = "IterableMatrix"), function(e1, e2) {
-  if (e1@transpose != e2@transpose) stop("Cannot add matrices with different internal transpose states.\nPlease use transpose_storage_order().")
-  if (e1@transpose) {
-    return(t(t(e1) + t(e2)))
-  }
-
-  assert_true(nrow(e1) == nrow(e2) && ncol(e1) == ncol(e2))
-
-  # If types are mismatched, default to double precision for both
-  type_e1 <- matrix_type(e1)
-  type_e2 <- matrix_type(e2)
-  if (type_e1 != type_e2 && type_e1 != "double") e1 <- convert_matrix_type(e1, "double")
-  if (type_e1 != type_e2 && type_e2 != "double") e2 <- convert_matrix_type(e2, "double")
-
-  dim <- c(nrow(e1), ncol(e1))
-  dimnames <- list(
-    merge_dimnames(rownames(e1), rownames(e2), "+", "row"),
-    merge_dimnames(colnames(e1), colnames(e2), "+", "column")
-  )
-  new("MatrixAddition", left = e1, right = e2, transpose = FALSE, dim = dim, dimnames = dimnames)
-})
 # Note: we skip numeric / IterableMatrix as it would result in a lot of infinities for dividing by 0.
 #' @describeIn IterableMatrix-methods Divide by a constant, or divide rows by a vector length nrow(mat)
 #' @examples
--- a/r/R/transforms.R
+++ b/r/R/transforms.R
@@ -752,6 +752,7 @@ setMethod("+", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
   e2 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e2, "double"))
   e2 + e1
 })
+# Note: setMethod("+", IterableMatrix, IterableMatrix) lives in matrix.R, alongside the MatrixAddition class definition.
 # Note: we skip numeric / IterableMatrix as it would result in a lot of infinities for dividing by 0.
 #' @describeIn IterableMatrix-methods Divide by a constant, or divide rows by a vector length nrow(mat)
 #' @examples
--- a/r/R/transforms.R
+++ b/r/R/transforms.R
@@ -740,7 +740,7 @@ setMethod("*", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
 #' mat + 1:nrow(mat)
 #' 
 #' ## Element-wise addition of two IterableMatrix objects
-#' mat1 + mat2
+#' mat + (mat * 2)
 #'
 setMethod("+", signature(e1 = "IterableMatrix", e2 = "numeric"), function(e1, e2) {
   if (all(e2 == 0)) return(e1)
@@ -761,10 +761,10 @@ setMethod("+", signature(e1 = "IterableMatrix", e2 = "IterableMatrix"), function
   assert_true(nrow(e1) == nrow(e2) && ncol(e1) == ncol(e2))
 
   # If types are mismatched, default to double precision for both
-  type_x <- matrix_type(e1)
-  type_y <- matrix_type(e2)
-  if (type_x != type_y && type_x != "double") e1 <- convert_matrix_type(e1, "double")
-  if (type_x != type_y && type_y != "double") e2 <- convert_matrix_type(e2, "double")
+  type_e1 <- matrix_type(e1)
+  type_e2 <- matrix_type(e2)
+  if (type_e1 != type_e2 && type_e1 != "double") e1 <- convert_matrix_type(e1, "double")
+  if (type_e1 != type_e2 && type_e2 != "double") e2 <- convert_matrix_type(e2, "double")
 
   dim <- c(nrow(e1), ncol(e1))
   dimnames <- list(
--- a/r/R/transforms.R
+++ b/r/R/transforms.R
@@ -727,7 +727,8 @@ setMethod("*", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
   e2 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e2, "double"))
   e2 * e1
 })
-#' @describeIn IterableMatrix-methods Add a constant, or row-wise addition with a vector length nrow(mat)
+#' @describeIn IterableMatrix-methods Add a constant, row-wise addition with a vector length nrow(mat),
+#'   or element-wise addition of two IterableMatrix objects
 #' @examples
 #' #######################################################################
 #' ## `e1 + e2` example
@@ -738,7 +739,9 @@ setMethod("*", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
 #' ## Adding row-wise by a vector of length `nrow(mat)`
 #' mat + 1:nrow(mat)
 #' 
-#' 
+#' ## Element-wise addition of two IterableMatrix objects
+#' mat1 + mat2
+#'
 setMethod("+", signature(e1 = "IterableMatrix", e2 = "numeric"), function(e1, e2) {
   if (all(e2 == 0)) return(e1)
   e1 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e1, "double"))
@@ -748,6 +751,27 @@ setMethod("+", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
   if (all(e1 == 0)) return(e2)
   e2 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e2, "double"))
   e2 + e1
+})
+setMethod("+", signature(e1 = "IterableMatrix", e2 = "IterableMatrix"), function(e1, e2) {
+  if (e1@transpose != e2@transpose) stop("Cannot add matrices with different internal transpose states.\nPlease use transpose_storage_order().")
+  if (e1@transpose) {
+    return(t(t(e1) + t(e2)))
+  }
+
+  assert_true(nrow(e1) == nrow(e2) && ncol(e1) == ncol(e2))
+
+  # If types are mismatched, default to double precision for both
+  type_x <- matrix_type(e1)
+  type_y <- matrix_type(e2)
+  if (type_x != type_y && type_x != "double") e1 <- convert_matrix_type(e1, "double")
+  if (type_x != type_y && type_y != "double") e2 <- convert_matrix_type(e2, "double")
+
+  dim <- c(nrow(e1), ncol(e1))
+  dimnames <- list(
+    merge_dimnames(rownames(e1), rownames(e2), "+", "row"),
+    merge_dimnames(colnames(e1), colnames(e2), "+", "column")
+  )
+  new("MatrixAddition", left = e1, right = e2, transpose = FALSE, dim = dim, dimnames = dimnames)
 })
 # Note: we skip numeric / IterableMatrix as it would result in a lot of infinities for dividing by 0.
 #' @describeIn IterableMatrix-methods Divide by a constant, or divide rows by a vector length nrow(mat)
--- a/r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
+++ b/r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
@@ -185,6 +185,18 @@ template <typename T> class MatrixAddition : public MatrixLoader<T> {
         return left.vecMultiplyLeft(v, user_interrupt) +
                right.vecMultiplyLeft(v, user_interrupt);
     }
+    Eigen::MatrixXd denseMultiplyRight(
+        const Eigen::Map<Eigen::MatrixXd> B, std::atomic<bool> *user_interrupt = NULL
+    ) override {
+        return left.denseMultiplyRight(B, user_interrupt) +
+               right.denseMultiplyRight(B, user_interrupt);
+    }
+    Eigen::MatrixXd denseMultiplyLeft(
+        const Eigen::Map<Eigen::MatrixXd> B, std::atomic<bool> *user_interrupt = NULL
+    ) override {
+        return left.denseMultiplyLeft(B, user_interrupt) +
+               right.denseMultiplyLeft(B, user_interrupt);
+    }
     // Linearity of sums (distribution over addition):
     //     rowSums(A+B) = rowSums(A) + rowSums(B) 
     //     colSums(A+B) = colSums(A) + colSums(B) 
--- a/r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
+++ b/r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
@@ -1,0 +1,153 @@
+// Copyright 2026 BPCells contributors
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// https://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or https://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#pragma once
+
+#include "MatrixIterator.h"
+
+namespace BPCells {
+
+// Element-wise addition of two input matrices with matching dimensions.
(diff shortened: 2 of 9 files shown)


bnprks:
Initially this looks pretty good! I think you mostly found all the useful helper functions, though I'll do a closer pass to look for any minor things that are missing.

One discussion I'd like to have: what are your thoughts on the main two algorithm options for doing the addition? As I see it, options are:
1. Calculate per-column sums as dense vectors (what you've done so far)
2. Assuming non-zero values within a column are stored in order by row, do something like a [2-way merge](https://en.wikipedia.org/wiki/Merge_algorithm) to create a sparse vector of the per-column sum (list of row idx + sum value).  
    - There's an OrderRows helper class that can guarantee sorted order by buffering + sorting as needed. Most inputs will start with row ordering, so it's usually just the cost of buffering, sometimes the cost of sorting (e.g. if there is a re-ordering matrix selection applied).

I'm not 100% sure which one makes more sense for this use-case, but I'd like to make sure that we've done some napkin math to think through the efficiency comparison between the algorithms.

erliuu:
Thanks for the suggestions!

Assuming sorted, the 2-way merge method does seem like a more feasible method (especially helpful with OrderRows). My understanding is that method 1 time scales based on the dimensions of the matrix (re-zeroes and sweeps n-rows buffer for each column). While method 2 only scales with the total number of non-zero elements. In the un-sorted case, OrderRows just uses radix sort which scales linearly.

Both cases put method 2 at `O(n)` while method 1 scales at `O(cols * rows + n)`, which should be strict asymptotic improvement. Please let me know if I made any incorrect assumptions here, or else I'm happy to draft up an implementation for it.

bnprks:
Yeah, that matches with my rough analysis, so I think it's worth implementing the 2-way merge approach if you're up for it. For the sparse matrices we work on, the non-zero fraction tends to be in the range of 0.01-0.05 for RNA (see fig S2 e+f in this [manuscript](https://www.biorxiv.org/content/10.1101/2025.03.27.645853v1.full.pdf)), so I often approximate n = 0.05 * cols * rows. That means we're often talking about a difference in constants in practice, though I still think it's likely to come out ahead for method 2 (more compact memory accesses, and I think possible to write mostly branch-free). In the event that matrices become very sparse, method 2 is likely to get a larger advantage as you point out with the asymptotic analysis. If you save a copy of method 1, you can always run a little performance comparison between the two to double-check on a test dataset.

erliuu:
@bnprks I've added an implementation for 2-way merge and a couple of tests to account for the merge behaviour. I just noticed you mentioned implementing this branch-free (which this is not). I'm not the most familiar with it, but happy to take a deeper look.

I'll run some quick performance comparison on the current branch then proceed with improvements.

bnprks:
Hi @erliuu, thanks, looks pretty good! Branch-free is ultimately a bit of reliance on the compiler, but the two tricks that seem to convince it most of the time is using ternary operators (`cond ? true_val : false_val`) and some boolean math on indices. For an example of this style, see [this bit](https://github.com/bnprks/BPCells/blob/28759cdd512578b6cbe549e226e1cd52a2d2308c/r/src/bpcells-cpp/matrixIterators/MatrixAccumulators.h#L59-L65) of MatrixAccumulators.h, which is "compacting" adjacent values that have matching row+col indices. Notice it still has one branch in the for loop, but that is highly predictable since it is not dependent on the data in the loop. (The main target is getting rid of *unpredictable* branches). It's not a perfect analogue for what you're doing here, but hopefully it gives some idea. (Another example is `InsertionIterator::nextInsertion()`) Happy to chat further if needed -- it's usually not that complicated to transform the if statements like you wrote to be branch free, but it might take some getting used to at first

I'll also see if I can dig up the instructions I made once on how to double-check what assembly code is generated for a function, since it's the real way to double-check if your code is compiled as branch free.

Other than that, there's just a couple testing details I'll need to double check to make sure that all the various edge cases about dimnames and subsetting are being exercised in testing. I think your code is correct since it's using all the right helpers, just trying to remember if there's a standard test that covers these details. There is at least one catch-all test `"Generic methods work"` which should be added to every time a new operator is added.

bnprks:
Found the link for checking generated assembly: https://github.com/bnprks/BPCells/wiki/Tools-and--Development-workflows#checking-generated-assembly-of-c-functions

r/R/transforms.R
```diff
@@ -749,6 +752,27 @@ setMethod("+", signature(e1 = "numeric", e2 = "IterableMatrix"), function(e1, e2
   e2 <- wrapMatrix("TransformScaleShift", convert_matrix_type(e2, "double"))
   e2 + e1
 })
+setMethod("+", signature(e1 = "IterableMatrix", e2 = "IterableMatrix"), function(e1, e2) {
```
bnprks: Nit: For consistency with `MatrixMultiply`, I would put this definition in `matrix.R` next to the `MatrixAddition` definition.

Discussion topic: I think we want to provide `-` if we provide `+`, and I lean towards doing that in R as `left + -1 * right`. Although a native C++ op is probably a bit better perf, the multiply + add option should still be reasonable as the multiply is pretty lightweight. I agree with not providing users a function if it will have bit performance pitfalls (e.g. we bail if transpose doesn't match rather than calling transpose_storage_order() automatically), but I think the perf will be fine for the R-only subtraction implementation.

On the other hand, it might be possible to use a template to have C++ matrix subtraction enabled with most code shared with C++ matrix addition. If you want to go that route I'm fine with a follow-up PR to add it in C++, but otherwise I think we can just add the R version now for a couple extra lines.

r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
```diff
@@ -0,0 +1,209 @@
...
+
+        row_buffer.resize(load_size);
+        val_buffer.resize(load_size);
+    }
+
+    uint32_t rows() const override { return left.rows(); }
+    uint32_t cols() const override { return left.cols(); }
+
+    const char *rowNames(uint32_t row) override { return left.rowNames(row); }
+    const char *colNames(uint32_t col) override { return left.colNames(col); }
```
bnprks, line 63: [No change needed; just mentioning] I'm realizing there's a slight conceptual mismatch between the C++ and R handling of dimnames here, as R will take names from `right` if the names on `left` are null but C++ will not. In the R library, dimnames really just reads from R vectors and I don't think this C++ will even get called when used from R. Fine to leave as is since rbind and cbind have the exact same issue, just something to keep in mind for e.g. a python port.

r/tests/testthat/test-matrix_transforms.R
```diff
@@ -8,7 +8,7 @@
 
 generate_sparse_matrix <- function(nrow, ncol, fraction_nonzero = 0.5, max_val = 10) {
   m <- matrix(rbinom(nrow * ncol, 1, fraction_nonzero) * sample.int(max_val, nrow * ncol, replace = TRUE), nrow = nrow)
-  as(m, "dgCMatrix")
+  as(Matrix::Matrix(m, sparse = TRUE), "dgCMatrix")
```
bnprks: question: did the `Matrix` package start spitting out warnings on `as(m, "dgCMatrix")`? I'm assuming that's why you needed to change this

r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
```diff
@@ -0,0 +1,209 @@
...
+                    r_idx++;
+                } else {
+                    T sum = left.valData()[l_idx] + right.valData()[r_idx];
+                    if (sum != 0) {
+                        row_buffer[loaded] = lr;
+                        val_buffer[loaded] = sum;
+                        loaded++;
+                    }
+                    l_idx++;
+                    r_idx++;
```
bnprks: This is the spot I would focus on trying to reduce unpredictable branches. For specific tactics, I would suggest:
1. Make this section handle values in bulk while left and right both have data, up to max load (yes a branch for the loop, but will come up true 100s of times in a row so easy to predict)
2. Use index-boolean addition and ternary operators to remove the need for if/else inside the loop. 
3. Leave the calls to `refill` outside the loop

Because we're usually loading in batches of 1024, this means the performance-sensitive inner loop will be fast and we mostly don't have to worry about minimizing branches elsewhere that happen once-per-batch rather than once-per-value.

r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
```diff
@@ -0,0 +1,209 @@
...
+
+    // Number of loaded entries available
+    uint32_t capacity() const override { return loaded; }
+
+    // Pointers to the loaded entries
+    uint32_t *rowData() override { return row_buffer.data(); }
+    T *valData() override { return val_buffer.data(); }
+
+    // MATH OPERATIONS: utilize distributive property that (A+B)*v = A*v + B*v
+    Eigen::VectorXd vecMultiplyRight(
```
bnprks, line 176: It's also worth implementing denseMultiplyRight and denseMultiplyLeft

bnprks (commented):
I just got a chance to review everything -- looks really good overall! I have a couple small comments/suggestions, including a note about where to focus on writing things branch-free. 

One other small suggestion I mentioned elsewhere is to add matrix addition to the `"Generic methods work"` test, which just requires adding one line with some way to use addition without modifying a matrix result (e.g. `0.5 * (mi + mi)`).

r/R/transforms.R
erliuu: Perhaps a minor readability issue: `MatrixMultiply` uses a distinct operator `%*%` compared to its matrix-scalar operator `*`. Meanwhile, `MatrixAddition` re-uses the same operator `+` for both. Since the `@describeIn` and `@examples` blocks show matrix-matrix addition, a developer reading transforms.R may expect the corresponding `setMethod` nearby. I'll add a comment to indicate where it can be found: 

```
# Note: setMethod("+", IterableMatrix, IterableMatrix) lives in matrix.R, alongside the MatrixAddition class definition.
```

erliuu: I'm happy with moving forward on the R-only `-` for now, and following up with a native C++ op in the future. 👍

bnprks: Having the extra code comment signpost sounds good. Then we get all the benefits of having `MatrixAddition`-related R code in a contiguous block in matrix.R but it's also clear where to look for more source code related to the documentation comment in transforms.R

erliuu:
Hey @bnprks,

Thanks for the feedback and suggestions. I've implemented your suggested changes. I just want to check that my branch-minimized implementation aligns with your comments about that section (and for my own understanding):

> 1. Make this section handle values in bulk while left and right both have data, up to max load (yes a branch for the loop, but will come up true 100s of times in a row so easy to predict)

My understanding is that I shouldn't be trying to make the whole check branchfree, but reduce unpredictable branches in the inner loop. This meant that I could keep the inner loop itself branched as its still highly predictable (one exit branch per `max_load_size` iterations).

```
while (loaded < max_load_size && l_idx < l_cap && r_idx < r_cap) {
```

> 2. Use index-boolean addition and ternary operators to remove the need for if/else inside the loop.

Claude gave me a good tip with the index checks and addition (`<=` and `>=` instead of additional `==` check) and I worked something like this out with ternary operators.

```
bool l_pick = lr_v <= rr_v;
bool r_pick = lr_v >= rr_v;
uint32_t out_row = l_pick ? lr_v : rr_v;
T out_val = l_pick ? (r_pick ? lv_v + rv_v : lv_v) : rv_v;
bool emit = (lr_v != rr_v) | (out_val != T(0));
row_buffer[loaded] = out_row;
val_buffer[loaded] = out_val;
loaded += emit;
l_idx += l_pick;
r_idx += r_pick;
```

> 3. Leave the calls to refill outside the loop

I popped the refill outside the inner-loop of the merge case. Is it worth it to move the refills out of all three cases (drain + merge) at the end of the outer loop? It would make the code cleaner, but my understanding is that both refill-checks would be conducted every outer loop, instead of only one side for the drain cases (Likely negligible in practice though).

erliuu:
I followed your instructions to generate the assembly for `000000000088eb20 W BPCells::MatrixAddition<unsigned int>::load()` and pulled what I thought was the relevant code for review with the python wrapper:

```
142	                while (loaded < max_load_size && l_idx < l_cap && r_idx < r_cap) {
   0x000000000088efe7 <+1223>:	mov    DWORD PTR [r15+0x148],ecx
   0x000000000088efee <+1230>:	cmp    ecx,DWORD PTR [r15+0x14c]
   0x000000000088eff5 <+1237>:	jae    88f050 <BPCells::MatrixAddition<unsigned int>::load()+1328>
   0x000000000088eff7 <+1239>:	cmp    eax,DWORD PTR [r15+0x154]
   0x000000000088effe <+1246>:	jae    88f0b8 <BPCells::MatrixAddition<unsigned int>::load()+1432>
   0x000000000088f004 <+1252>:	cmp    DWORD PTR [r15+0x15c],edx
   0x000000000088f00b <+1259>:	jbe    88f050 <BPCells::MatrixAddition<unsigned int>::load()+1328>
   0x000000000088f00d <+1261>:	add    rax,r8
   0x000000000088f010 <+1264>:	add    rdx,r14
143	                    uint32_t lr_v = lr[l_idx];
   0x000000000088f013 <+1267>:	mov    edi,DWORD PTR [r9+rax*4]
144	                    uint32_t rr_v = rr[r_idx];
   0x000000000088f017 <+1271>:	mov    esi,DWORD PTR [r13+rdx*4+0x0]
145	                    T lv_v = lv[l_idx];
   0x000000000088f01c <+1276>:	mov    ebp,DWORD PTR [rbx+rax*4]
146	                    T rv_v = rv[r_idx];
   0x000000000088f01f <+1279>:	mov    r10d,DWORD PTR [r12+rdx*4]
147	
148	                    bool l_pick = lr_v <= rr_v;
   0x000000000088f023 <+1283>:	cmp    edi,esi
   0x000000000088f025 <+1285>:	mov    r11d,esi
   0x000000000088f028 <+1288>:	cmovbe r11d,edi
   0x000000000088f02c <+1292>:	setbe  al
149	                    bool r_pick = lr_v >= rr_v;
   0x000000000088f02f <+1295>:	setae  dl
150	                    uint32_t out_row = l_pick ? lr_v : rr_v;
151	                    T out_val = l_pick ? (r_pick ? lv_v + rv_v : lv_v) : rv_v;
   0x000000000088f032 <+1298>:	ja     88ef93 <BPCells::MatrixAddition<unsigned int>::load()+1139>
   0x000000000088f038 <+1304>:	jae    88ef90 <BPCells::MatrixAddition<unsigned int>::load()+1136>
   0x000000000088f03e <+1310>:	mov    r10d,ebp
   0x000000000088f041 <+1313>:	jmp    88ef93 <BPCells::MatrixAddition<unsigned int>::load()+1139>
   0x000000000088f046 <+1318>:	cs nop WORD PTR [rax+rax*1+0x0]
```

I can't say I know how to read assembly well, but my understanding is that we are trying to minimize the `j*` conditional jumps in the main body of the inner-loop (merge-case). Lines 143-150 appear to show a successful branchfree implementation of the changes, but Line 151 seems to show that the output value ternary still compiled branched. Please let me know if I am looking at this correctly, and what next steps might be. Thanks!

r/src/bpcells-cpp/matrixIterators/MatrixAddition.h
```diff
@@ -0,0 +1,221 @@
...
+                    uint32_t rr_v = rr[r_idx];
+                    T lv_v = lv[l_idx];
+                    T rv_v = rv[r_idx];
+
+                    bool l_pick = lr_v <= rr_v;
+                    bool r_pick = lr_v >= rr_v;
+                    uint32_t out_row = l_pick ? lr_v : rr_v;
+                    T out_val = l_pick ? (r_pick ? lv_v + rv_v : lv_v) : rv_v;
+
+                    bool emit = (lr_v != rr_v) | (out_val != T(0));
```
bnprks:
```suggestion
                    bool emit = out_val != T(0);
```
I don't think the `lr_v != rr_v` is needed? I can't think of a case where the row index comparison would be needed

r/tests/testthat/test-matrix_utils.R
```diff
@@ -936,7 +938,7 @@ test_that("Generic methods work", {
     expect_identical(rowSums(trans), rowSums(m))
     expect_identical(colSums(trans), colSums(m))
 
-    if (i %in% c("shift_scale_1", "shift_scale_2")) {
+    if (i %in% c("shift_scale_1", "shift_scale_2", "add", "subtract")) {
```
bnprks:
```suggestion
    if (i %in% c("shift_scale_1", "shift_scale_2")) {
```
The reason for this if statement is that the shift transforms can output explicit zeros (e.g. mat + 1 - 1). I don't think we need to use it for the new cases, since they should never output explicit zeros

bnprks (approved):
Nice job on the adjustments, just two small one-line suggestions on my end. 

> My understanding is that I shouldn't be trying to make the whole check branchfree, but reduce unpredictable branches in the inner loop. This meant that I could keep the inner loop itself branched as its still highly predictable (one exit branch per max_load_size iterations).

Yep, exactly right

> Is it worth it to move the refills out of all three cases (drain + merge) at the end of the outer loop?

Because it's out of the hot loop I think it probably doesn't matter and you can do whatever makes the code easiest to read/think about in your judgement

> Lines 143-150 appear to show a successful branchfree implementation of the changes, but Line 151 seems to show that the output value ternary still compiled branched. Please let me know if I am looking at this correctly, and what next steps might be.

I agree with your analysis that we only hit partially branch free. I checked a few tricks on my end and also wasn't successful getting the compiler to spit out branch-free code for `out_val`. (Tricks were 1. using clang instead of gcc 2. Trying `out_val = l_pick * lv_v + r_pick * rv_v`). It seems my intuition here was a bit off for what would produce branch-free code

I'd recommend taking the partial success and moving on, as further wrestling with the compiler would require (1) running actual timing tests/[profiling](https://github.com/bnprks/BPCells/wiki/Tools-and--Development-workflows#profiling-c-code) to get ground-truth performance data (2) trying some more variants of how to write the inner loop (maybe first merge values into the output buffer without adding, then do the addition/zero dropping in a second loop? Maybe there are other slight tweaks possible?). You can develop intuitions over time for what might work and what might not, but at the end of the day there's always an element of guess and check. 

I don't *think* that this matrix addition is likely to be a common bottleneck, hence why I think it might be easiest to just move on. If you're having fun with this as a chance to learn more about inspecting assembly/profiling code then go for it, otherwise I think this code is basically ready to merge

r/tests/testthat/test-matrix_utils.R
erliuu: Ah got it. I originally thought we might want to keep explicit zeros (hence condition in line 153 in MatrixAddition.h), but this definitely makes more sense. Applying both fixes.

PR #337, opened 2026-05-06
[r] Support graph list inputs in clustering

koefoeden:
## Context
I ran into a large transient memory spike when clustering a shared-nearest-neighbor graph built from about 463k cells and 20 embedding dimensions. The KNN search completed, but memory climbed sharply during graph clustering. Smaller benchmarks pointed to the `igraph::graph_from_adjacency_matrix()` conversion inside the current clustering path as the expensive step, rather than KNN search, SNN construction, or the final membership vector.

Representative peak RSS measurements from the same graph construction/clustering path were:

| Input / stage | Peak RSS |
|---|---:|
| 20k cells, KNN only | ~0.59-0.66 GiB |
| 20k cells, KNN + SNN | ~0.68-0.70 GiB |
| 20k cells, KNN + SNN + Leiden through sparse matrix | ~5.98-6.00 GiB |
| 30k cells, sparse-matrix clustering route | ~12.58 GiB |
| 30k cells, direct edge-list route | ~0.87 GiB |
| 50k cells, sparse-matrix clustering route | ~42.9 GiB |
| 50k cells, direct edge-list route | ~1.04 GiB |

A separate 20k-cell equivalence check produced identical Leiden memberships between the sparse matrix route and the direct edge-list route. This PR upstreams that lower-memory route for BPCells graph-list outputs while leaving the existing sparse matrix behavior unchanged.

## Summary
- Allow `cluster_graph_leiden()` and `cluster_graph_louvain()` to accept graph-list outputs from `knn_to_snn_graph(return_type = "list")` and `knn_to_geodesic_graph(return_type = "list")`.
- Build igraph objects directly from the edge list for list inputs, while keeping the existing sparse matrix path unchanged.
- Convert BPCells 0-based edge-list indices to igraph 1-based vertex IDs internally.
- Ignore self-loops for list inputs to match the existing `diag = FALSE` matrix behavior.
- Add regression coverage that matrix and graph-list inputs produce identical Leiden and Louvain memberships on the same graph.

## Testing

```r
testthat::test_file("tests/testthat/test-clustering.R")
```

Result: 13 passed; 2 skipped because optional packages `RcppAnnoy` and `RcppHNSW` were not installed in the test environment.

The new regression assertions are in the existing `igraph clustering doesn't crash` test and did run. The two skipped tests are separate optional ANN tests guarded by `skip_if_not_installed("RcppAnnoy")` and `skip_if_not_installed("RcppHNSW")`; I did not install those optional packages for this focused test run.

## Disclosure
GPT-5.5 was utilized to help draft this PR and implementation.

### DIFF
--- a/r/R/clustering.R
+++ b/r/R/clustering.R
@@ -234,13 +234,13 @@ knn_to_geodesic_graph <- function(knn, return_type = c("matrix", "list"), thread
   return(res)
 }
 
-#' Cluster an adjacency matrix
+#' Cluster an adjacency matrix or graph edge list
 #' @rdname cluster_graph
 #' @details **cluster_graph_leiden**: Leiden clustering algorithm `igraph::cluster_leiden()`. 
 #'    Note that when using `objective_function = "CPM"` the number of clusters empirically scales with `cells * resolution`,
 #'    so 1e-3 is a good resolution for 10k cells, but 1M cells is better with a 1e-5 resolution. A resolution of 1 is a 
 #'    good default when `objective_function = "modularity"` per the default.
-#' @param mat Symmetric adjacency matrix (dgCMatrix) output from e.g. `knn_to_snn_graph()` or `knn_to_geodesic_graph()`. Only the lower triangle is used.
+#' @param mat Symmetric adjacency matrix (dgCMatrix) or graph list output from e.g. `knn_to_snn_graph()` or `knn_to_geodesic_graph()`. Only the lower triangle is used for matrix inputs.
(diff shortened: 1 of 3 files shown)


koefoeden:
Changed to draft status due to possible regression detected. Will update!

koefoeden:
No regression afterall. It was a high number of single-tons in my case that resulted in >100 "distinct" clusters. I implemented my own helper to discard clusters below a given number of barcodes.

PR #339, opened 2026-06-13
[r][cpp] add Bernoulli Sampling

erliuu:
Hi @bnprks,

Been sitting on this for a while, but thought I would finally PR a useful feature I've been using downstream locally with IterableMatrices (particularly for doublet simulation).

### Description

This PR adds a [Bernoulli sampling](https://en.wikipedia.org/wiki/Bernoulli_sampling) function for an `IterableMatrix` to keep each non-zero entry independently based on provided probability `prob`. So for `prob=0.3`, we expect to keep approximately 30% of non-zeros. This should require no materialization or additional allocation.

- Adds a `BernoulliSample` class that performs the streaming keep/drop filter based off of the `FilterZeros` transform.

### Details

- The sampling is designed to be deterministic. Keep/drop choice is a `splitmix64` hash of seed, column, and row.
- As the column index is taken after subset, by design, selecting the same input column multiple times will still give independent sampling not identical copies. (e.g. for doublet simulation, when `MatrixSubset` selects the same column more than once).
- Works for `uint32_t`, `float`, and `double` matrices

### Tests

- Add functional tests in `test-matrix_transforms.R`: `prob=1` no-op, keep percentage within tolerance, reproducibility, independent draws on duplicate subset columns, matrix-type invariance behaviour.

Happy to clarify any questions you might have, thanks!


r/tests/testthat/test-matrix_transforms.R
```diff
@@ -418,3 +418,62 @@ test_that("matrix subtraction works", {
...
+    # Test hash keying to output columns (duplicated subset columns get independent draws)
+    col1 <- ceiling(ncol(bp) / 2)
+    dup_subset <- methods::new(
+        "MatrixSubset",
+        matrix = bp,
+        col_selection = rep(as.integer(col1), 3),
+        dim = c(nrow(bp), 3L),
+        dimnames = list(rownames(bp), NULL),
+        zero_dims = c(FALSE, FALSE)
+    )
```
bnprks: This seems like a slightly weird subtest, as my recollection is that our subset op explicitly throws an error if you try to duplicate columns/rows using it.

r/tests/testthat/test-matrix_transforms.R
```diff
@@ -418,3 +418,62 @@ test_that("matrix subtraction works", {
...
+    nnz_in <- length(m_big@x)
+    out <- as(bernoulli_sample(bp_big, 0.5, 17L), "dgCMatrix")
+    nnz_out <- length(out@x)
+    sigma <- sqrt(0.25 * nnz_in)
+    expect_lt(abs(nnz_out - 0.5 * nnz_in), 4 * sigma)
+
+    # Reproduciblity: same seed materializes identical output
+    y1 <- as(bernoulli_sample(bp, 0.3, 42L), "dgCMatrix")
+    y2 <- as(bernoulli_sample(bp, 0.3, 42L), "dgCMatrix")
+    expect_identical(y1, y2)
```
bnprks, line 442: As per overall comments, might want to add a reproducibility check when using `transpose_storage_order()`

bnprks (commented):
Hi @erliuu,
1. At the R level, I agree using `@transpose` flag is the way to go. I was more thinking at the C++ level having a second class might be convenient, but I think any option is fine that keeps the boolean check on transpose out of the hot `for` loop in `load()`. Happy to have it in this PR if it'll be fast to add in
2. With some further reading, I think splitmix64 seems to be fairly well regarded so no need to switch to pcg. In particular, splitmix64 seems to fare well in a benchmark by Daniel Lemire who I trust: https://github.com/lemire/testingRNG. I think the size of the state is likely to be similar between pcg and splitmix64, but it seems splitmix64 is a bit faster and as a bonus has a very simple implementation.

After taking a closer look at splitmix64, I'm not quite sure that it's being called correctly:

3. In this PR's `splitmix64`, the function is fully stateless, but in [Lemire's implementation](https://github.com/lemire/testingRNG/blob/master/source/splitmix64.h) and the [Java stdlib](https://github.com/openjdk/jdk/blob/7e356c4eb5fdea3de0db6496707f21ae8dbe29b6/src/java.base/share/classes/java/util/SplittableRandom.java), there is a `uint64_t` as state. Also, the reference implementations don't do any `^` ops on the state/arguments like happens in this PR. I think that our `splitmix64` is therefore slightly wrong (or at least nonstandard), and since getting a good RNG can be subtle, I think it's best to try to match how trustworthy sources do it.
   
   Because the state manipulation is quite simple (adding a constant), I wonder if we can use something like the `splitmix64_stateless(seed, offset)` from the Lemire code, and set `offset = col * nrow + row` or `offset = row * ncol + col` depending on the transpose state. (It need not be this exact scheme, just something that maps each matrix entry to a unique element from the random stream)
4. One possible naming style suggestion: What if the R function was called `sample_bernoulli()` instead of `bernoulli_sample()`? Then it leaves room for future `sample_poisson` or `sample_binomial` functions, following the prefix style of our `knn_*`, `plot_*`, `cluster_*` functions. I know it reads a bit less naturally, though I've always liked how it makes alphabetical lists group nicely and easier discoverability with tab completion after typing, e.g. `cluster_<tab>` (optional; curious for your thoughts)

